PR #132, opened 2014-09-18
OPTION_LIBICAL_COMPATIBLE

mirabilos:
https://github.com/fruux/sabre-vobject/issues/129

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -18,6 +18,8 @@ ChangeLog
 * #136: Validator rejects *all* control characters. We were missing a few.
 * #133: Splitter objects will throw exceptions when receiving incompatible
   objects.
+* #129: Added: OPTION_LIBICAL_COMPATIBLE parses libical-specific and
+  otherwise invalid escapes in TEXT in the same manner as libical does
 
 3.3.1 (2014-08-18)
 ------------------
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -398,7 +398,7 @@ class MimeDir extends Parser {
         if (strtoupper($propObj['ENCODING']) === 'QUOTED-PRINTABLE') {
             $propObj->setQuotedPrintableValue($this->extractQuotedPrintableValue());
         } else {
-            $propObj->setRawMimeDirValue($property['value']);
+            $propObj->setRawMimeDirValue($property['value'], $this->options);
         }
 
         return $propObj;
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -468,7 +468,10 @@ class MimeDir extends Parser {
      */
     static public function unescapeValue($input, $delimiter = ';', $options = 0) {
 
-        $regex = '#  (?: (\\\\ (?: \\\\ | N | n | ; | , ) )';
+        $regex = '#  (?: (\\\\ (?: \\\\ | N | n | ; | ,';
+        if ($options & self::OPTION_LIBICAL_COMPATIBLE)
+            $regex .= ' | t | r | b | f | "';
+        $regex .= ' ) )';
         if ($delimiter) {
             $regex .= ' | (' . $delimiter . ')';
         }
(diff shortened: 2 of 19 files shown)


Hywan:
ping?

mirabilos:
is there anything I need to do, or is the ping not for me?

Hywan:
Is the PR ready for a review?

evert:
The underlying issue (https://github.com/libical/libical/issues/74) has since been fixed, and @mirabilos also pointed out that that fix works for him.

I'm wondering if this is still important. It was a bit of an edge-case to begin with (specifically a workaround for people using a library with a bug, and feeding it data that iCalendar should not support).

We've so far haven't heard of other people with this issue, so I'm tempted to just close it.

mirabilos:
@evert the fix in libical works for me, but this change to vobject is still absolutely required. There are lots of ical files/mails created by older libical versions out there, plus, the next Debian stable that will carry the libical fix is expected to be released around 2018, and the next Ubuntu LTS with it also only in 2016, which means that users will carry it around for a long time (probably around 2019, when the currently-prepared Debian stable will have its oldstable end of life).

Yes, it is an edge case, but it is an actual interoperability problem. Be conservative in what you send but liberal in what you accept.

Hywan:
@mirabilos What about a warning? Would it be a good compromise? We can detect the version of libical or add a mention in the `README` or `UPDATE` files?

Hywan:
(@mirabilos I am just asking, feel free to say no :hankey:!)

mirabilos:
A warning is not a compromise. Detecting the version of libical is impossible because, by the very nature of this, it’s a networked protocol: the ical files and iTip/iMip messages are generated on a *remote* system, and then processed on another, possibly by vobject.

All this pull request does is to add a new flag in which libical-bug-compatible mode can be chosen, it does not change the default, so applications which do wish to import these files/messages can do so. It’s an absolute requirement for the Kolab-ActiveSync-bridge application I’m working on.

evert:
@mirabilos it's an absolute requirement, or it _was_ an absolute requirement before the fix in libical? The distinction is important :)

One other possibility for yourself, would be to pre-process the files. I think you can determine with high certainty that any backslash followed by an invalid character can be stripped.

It would even be possible to do this with a php stream filter, so you don't need to load the entire string into memory.

There is still the issue here that this is a pretty involved change, fixing an issue for (so far) 1 person, if that person:

1. uses an outdated version of libical
2. feeds libical data that should never have made it into an iCalendar object in the first place.

The question I need to ask myself is, is this situation common enough to warrant a workaround in vobject.

mirabilos:
It is still an absolute requirement, because data created by the broken libical versions will be around for the next 8-10 years, in my estimate (since the broken libical versions will stay to be in use for the next five years at least).

Why are you suggesting things like filters now, when we have a perfectly fine, working solution which cost me several hours of time (paid by the customer, but still)?

evert:
I totally understand that that's frustrating. One thing I didn't realize before the patch, was that I was under the assumption that it would only touch Parser\MimeDir. It got a lot more ugly than that.

I'm trying to objectively decide whether or not it's worth it. The hours you spent is 'sunk cost', but adding a feature is baggage and we'll spend time on this for years in the future as we're responsible for maintaining this.

And it should also be noted that adding 'sanitizing' the data before giving it to vobject was something that I've suggested very early on.

Would it be possible to refactor the patch, so there's a 1-time filter per 'input line', so by the time the value gets to the properties, the data is already sanitized? That would make it a lot easier to maintain and would reduce the entire feature to a single file.

Hywan:
@evert +1. I also like the idea of having filters per “3rd party”.

@mirabilos You' re saying libical will be around for the next 10 years. You forget that at this time, sabre will rulz the worldz :metal:!

mirabilos:
@Hywan I’m saying libical will be around for a couple of years mostly based on the expected lifetime of Debian “jessie”, which is frozen and to be released Really Soon Now™ (possibly in time for FOSDEM).

Also, I’d be interested to know how you’d patch KDEPIM/Kontact (C++ code), which currently calls out to libical (C code), to use Sabre (PHP code) instead ☻☺ No offence intended.

mirabilos:
@evert you cannot simply filter the input because you have to be aware of the precise point in which you are in the input string (e.g. you can’t just remove “\t” because that would also hit “\\t”). Furthermore, there’s MIME encoding to consider (which is also currently done by Sabre), and a filter would have to sit between MIME (Sabre) and parsing (Sabre), so Sabre *is* the obvious place for this.

I’m sad to hear you consider not carrying the patch. I will have to ship a modified version along with the software then.

evert:
You could definitely figure out, based on context, what the intent is.. just as you are doing right now.

If you see a `\` in a line, you check the character that follows. If it's a `t`, you drop remove the entire `\t`, if it's a `\`, you ignore it. This is not that different from how it works now.

The biggest difference is that it would reduce the size of the patch quite a bit.

mirabilos:
Evert Pot dixit:

>realize before the patch, was that I was under the assumption that it
>would only touch Parser\MimeDir. It got a lot more ugly than that.

That’s mostly due to your API, and your recommendation that it be
a parser option.

All in all, it is not very ugly: the first of the commits is
purely mechanical, the rest rather self-contained.

bye,
//mirabilos
-- 
„Cool, /usr/share/doc/mksh/examples/uhr.gz ist ja ein Grund,
mksh auf jedem System zu installieren.“
	-- XTaran auf der OpenRheinRuhr, ganz begeistert
(EN: “[…]uhr.gz is a reason to install mksh on every system.”)

evert:
I take the 'bye' as not being interested in looking for different solutions. If this issue turns out to affect other people in the future, we can reconsider and reopen this.

PR #138, opened 2014-09-19
fix convertToUTF8 and amend testValidateNonUTF8()

mirabilos:
here’s the remaining fixes

### DIFF
--- a/lib/StringUtil.php
+++ b/lib/StringUtil.php
@@ -43,26 +43,39 @@ class StringUtil {
      * @return string
      */
     static public function convertToUTF8($str) {
+        /*
+         * Unfortunately, mb_check_encoding is not reliable.
+         * But mb_convert_encoding can be used to convert
+         * from (presumed) UTF-8 to UTF-16 then back to UTF-8,
+         * and will replace invalid input with question marks.
+         * That means if previous and result are not equal,
+         * the input was not UTF-8, in which case we convert
+         * it as if it were ISO-8859-1; all 256 bytes are
+         * (after conversion) valid per RFC3629 §4. We need,
+         * for this to work, to temporarily change the internal
+         * encoding used by the mb_* functions, though.
+         */
+        $mb_encoding = mb_internal_encoding();
+        mb_internal_encoding("UTF-8");
 
-        $encoding = mb_detect_encoding($str , array('UTF-8','ISO-8859-1', 'WINDOWS-1252'), true);
-
-        switch($encoding) {
-            case 'ISO-8859-1' :
-                $newStr = utf8_encode($str);
-                break;
-            /* Unreachable code. Not sure yet how we can improve this
-             * situation.
-            case 'WINDOWS-1252' :
-                $newStr = iconv('cp1252', 'UTF-8', $str);
-                break;
-             */
-            default :
-                 $newStr = $str;
-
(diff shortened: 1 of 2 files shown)


lib/StringUtil.php
```diff
@@ -43,26 +43,39 @@ static public function isUTF8($str) {
...
-             * situation.
-            case 'WINDOWS-1252' :
-                $newStr = iconv('cp1252', 'UTF-8', $str);
-                break;
-             */
-            default :
-                 $newStr = $str;
-
+        /* coerce to string first */
+        $str = ''.$str;
```
Hywan: $str = (string) $str;

Hywan:
Actually, I don't understand what this PR does. @mirabilos, what are you trying to resolve?

mirabilos:
It correctly handles the cases of incoming latin1 and UTF-8 strings. (I could enhance it to support windows-1252 as well.) #156 which you merged drops this ability, if it is even correct at all.

Hywan:
Could you give me a usecase please? I wonder if we can find a simple solution?

mirabilos:
Simply spoken, the mb_{check,detect}_encoding functions do not always check the validity of the string in the specified encoding (which is what they should do). Converting between UTF-8 and UTF-16 and back is a good check (although I just looked at the PCRE-based one, which *should* also work), and converting a string *to* UTF-8 if it’s latin1 can be done with the mb_* conversion function more easily (utf8_encode is legacy, *and* travels through the entire XML stack). Heck, extending it to actually support windows-1252 would be easy.

Maybe this could be shorted by using the PCRE version. But when I submit something, I prefer solutions I know to work, from production code here.

Hywan:
@mirabilos I fixed the underscore issue with your comment.

The `preg_match` trick comes from [`Hoa\String`](https://github.com/hoaproject/String) which is a well-tested Unicode string library. So I tend to be confident with this solution. So it removes the needs of encoding to UTF-16 and then encoding back to UTF-8.

Then, is the goal also to remove control characters? Actually, they are valid UTF-8 characters but in [sabre/vobject we consider them as invalid](https://github.com/fruux/sabre-vobject/blob/b914cf931ddf2991022bc30cc00a8a2fef541a2d/lib/StringUtil.php#L22-L25). If yes, we can use a regular expression based on [`Hoa\String\String::getCharWidth`](https://github.com/hoaproject/String/blob/c3e510b9aaf685e9bba2dcddd7ee23850faddfe4/String.php#L864-L912) which counts the number of column needed to print a char. It takes care of the control characters. We can use this algorithm as a basis.

Note: I was reviewing your PR again. All the `mb_*` stuff is not necessary if we consider the `preg_match` trick efficient. Then, your last `return` statement combined with a `preg_replace` is valid.

mirabilos:
Hrm. I can’t comment in-detail enough right now, but you do raise some important points. Next time I can work on this project, I’ll have a look at all possible callers, and try to figure something out. Thanks for the links also.

Can we just leave this open but 'put back' for now, then?

Hywan:
Oops, forget about the “invalid” label, I mistaken. Now it's in “question”, so it's fine. We keep it open.

evert:
It's been a long time since the last activity here. If there are any problems here, drop me a line so we can reopen.

PR #160, opened 2014-11-13
Implement xCal support

Hywan:
Useful links:
- [RFC6321](https://tools.ietf.org/html/rfc6321), xCal: The XML Format for iCalendar.

To-do:
- [x] require sabre/xml,
  - [x] update `composer.json`,
  - [x] sabre/xml requires PHP5.4, hence update sabre/vobject (done by Composer?),
    - [x] if requires PHP5.4, we can use the [`JsonSerializable`](http://php.net/jsonserializable) everywhere (all top-classes having a `jsonSerialize` method),
    - [x] update code accordingly (e.g. short array syntax),
    - [ ] update documentation accordingly,
- [x] create the helper in `Sabre\VObject\Reader`,
- [x] add XML serializer helpers everywhere (e.g. in children of `Sabre\VObject\Node`),
- [x] timezones used by xCal and xCard are differents, must check the compatibility,
- [x] create the reader (parser):
  - [x] support properties:
    - [x] support all names,
    - [x] support all values (types of values):
      - [x] `binary`,
      - [x] `boolean`,
      - [x] `cal-address`,
      - [x] `date`,
      - [x] `date-time`,
      - [x] `duration`,
      - [x] `float`,
      - [x] `integer`,
      - [x] `period`,
        - [x] `start`,
        - [x] `duration`,
        - [x] `end`,
      - [x] `recur`,
        - [x] `until`,
        - [x] `count`,
        - [x] `interval`,
        - [x] `bysecond`,
        - [x] `byminute`,
        - [x] `byhour`,
        - [x] `byday`,
        - [x] `bymonthday`,
        - [x] `byday`,
        - [x] `byyearday`,
        - [x] `byweekno`,
        - [x] `bymonth`,
        - [x] `bysetpos`,
        - [x] `wkst`,
        - [x] `freq`,
      - [x] `text`,
      - [x] `time`,
      - [x] `uri`,
      - [x] `utc-offset`,
    - [x] support special cases:
      - [x] `categories`,
      - [x] `resources`,
      - [x] `freebusy`,
      - [x] `exdate`,
      - [x] `rdate`,
      - [x] `geo`,
        - [x] `latitude`,
        - [x] `longitude`,
      - [x] `request-status`,
        - [x] `code`,
        - [x] `description`,
    - [x] support all parameters:
      - [x] support the specific `value` iCalendar parameter,
  - [x] support all components:
    - [x] `vevent`,
    - [x] `vtodo`,
    - [x] `vjournal`,
    - [x] `vfreebusy`,
    - [x] `vtimezone`,
    - [x] `standard`,
    - [x] `daylight`,
    - [x] `valarm`,
  - [x] support XML extensions.

Code review:
- usage of iterators has been heighten (see https://github.com/fruux/sabre-vobject/pull/161),
- clean code (https://github.com/fruux/sabre-vobject/pull/162),
- better performance when possible (see https://github.com/fruux/sabre-vobject/pull/162).

Left apart for now:
- [x] [RFC6321, section 3.1. Pre-Processing](https://tools.ietf.org/html/rfc6321#section-3.1),
- [x] `LIBXML_*` options when reading or writing.

To fix:
- [x] `rdate`.

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -364,9 +364,7 @@ abstract class Property extends Node {
             foreach ($parameters as $parameter) {
 
                 $writer->startElement(strtolower($parameter->name));
-                    $writer->startElement('text');
-                        $parameter->xmlSerialize($writer);
-                    $writer->endElement();
+                $writer->writeElement('text', $parameter);
                 $writer->endElement();
 
             }
@@ -390,11 +388,7 @@ abstract class Property extends Node {
     protected function xmlSerializeValue(Xml\Writer $writer) {
 
         foreach ($this->getJsonValue() as $value) {
-
-            $writer->startElement(strtolower($this->getValueType()));
-                $writer->write($value);
-            $writer->endElement();
-
+            $writer->writeElement(strtolower($this->getValueType()), $value);
         }
 
     }
--- a/lib/Property/Float.php
+++ b/lib/Property/Float.php
@@ -127,13 +127,8 @@ class Float extends Property {
 
             $value = array_map('floatval', $this->getParts());
 
-            $writer->startElement('latitude');
-                $writer->write($value[0]);
-            $writer->endElement();
-
-            $writer->startElement('longitude');
-                $writer->write($value[1]);
-            $writer->endElement();
+            $writer->writeElement('latitude', $value[0]);
+            $writer->writeElement('longitude', $value[1]);
 
         }
         else {
--- a/lib/Property/ICalendar/Period.php
+++ b/lib/Property/ICalendar/Period.php
@@ -136,22 +136,15 @@ class Period extends Property {
     protected function xmlSerializeValue(Xml\Writer $writer) {
 
         $writer->startElement(strtolower($this->getValueType()));
-
         $value = $this->getJsonValue();
-
-        $writer->startElement('start');
-            $writer->write($value[0][0]);
-        $writer->endElement();
+        $writer->writeElement('start', $value[0][0]);
 
         if ($value[0][1][0] === 'P') {
-            $writer->startElement('duration');
+            $writer->writeElement('duration', $value[0][1]);
         }
         else {
-            $writer->startElement('end');
+            $writer->writeElement('end', $value[0][1]);
         }
-
-        $writer->write($value[0][1]);
-        $writer->endElement();
 
         $writer->endElement();
 
--- a/lib/Property/Text.php
+++ b/lib/Property/Text.php
@@ -304,20 +304,11 @@ class Text extends Property {
 
             $value = $this->getJsonValue();
 
-            $writer->startElement('code');
-                $writer->write($value[0][0]);
-            $writer->endElement();
-
-            $writer->startElement('description');
-                $writer->write($value[0][1]);
-            $writer->endElement();
+            $writer->writeElement('code', $value[0][0]);
+            $writer->writeElement('description', $value[0][1]);
 
             if (isset($value[0][2])) {
-
-                $writer->startElement('data');
-                    $writer->write($value[0][2]);
-                $writer->endElement();
-
+                $writer->writeElement('data', $value[0][2]);
             }
 
         }


Hywan:
Back

Hywan:
(`git rebase master`)

lib/Parser/XML.php
```diff
@@ -0,0 +1,264 @@
...
+                                $propertyParameters[static::getTagName($xmlParameter['name'])]
+                                    = $xmlParameter['value'][0]['value'];
+
+                            array_splice($xmlProperty['value'], $i, 1);
+
+                        }
+
+                        switch ($propertyName) {
+
+                            case 'dtstart':
```
evert: Curious why `dtstart`, `geo`, etc, all need special-casing.

Hywan: `geo`, `request-status` etc. are [special cases according to the RFC](http://tools.ietf.org/html/rfc6321#section-3.4.1).
`dstart` is a bug fix. I have to dig more about this one. See my comment here https://github.com/Hywan/sabre-vobject/commit/46ef57ceab8cb26fa89552da9a4dfb45fe1519e4#commitcomment-9279424.

Hywan:
All examples and usecases given in the RFC work!

Hywan:
36 to 62 assertions, same number of tests, several bugs found, with the reflexive equality asserter helper. Nice improvement!

Now, I have to check if the implementation of `<parameters>` is safe. Then, I will refactore my code a little bit to use the `XML\Writer` instance everywhere instead of computing arrays.

Hywan:
Almost finished. I have to make a final review to check all comments.

lib/Property.php
```diff
@@ -319,6 +319,77 @@ function jsonSerialize() {
...
     }
 
+    /**
+     * Returns the value, in the format it should be encoded for XML.
+     *
+     * This method must always return an array.
+     *
+     * @return array
+     */
+    function getXmlValue() {
```
Hywan: To clarify the discussion, we pass `Sabre\Xml\Writer` everywhere now.

lib/Property.php
```diff
@@ -319,6 +319,77 @@ function jsonSerialize() {
...
+
+    /**
+     * Sets the XML value, as it would appear in a xCard or xCal object.
+     *
+     * The value must always be an array.
+     *
+     * @param array $value
+     * @return void
+     */
+    function setXmlValue(array $value) {
```
Hywan: All `getXmlValue` methods have been removed.
All `xmlSerialize` methods now receive a `Sabre\Xml\Writer` object.
The `setXmlValue` are still based on array because the XML parser works with an array. These is 1 definition of `setXmlValue` and only 4 overwrites. I think it's fine enough.

lib/Parser/XML.php
Hywan: `dtstart` is no longer a special case (see https://github.com/fruux/sabre-vobject/commit/34fc363991bf242facff41da0b86c980ad00ac8b) and the algorithm entirely reflects the RFC now.

Hywan:
I checked all comments. Everything is solved. I am going to write few more tests but you can start the review :-). /cc @evert

Hywan:
I am adding the Relax NG Schema, in the compact form. libxml2 (so XMLReader) does not read the compact form, only the XML one. I am trying to convert it but not sure if it is useful. The goal is to provide the schema, and people can use the appropriate tool to read it. http://relaxng.org/ lists some of them.

evert:
So sad libxml doesn't relax NG properly! I was _so_ looking forward to start using it!

composer.json
```diff
@@ -6,7 +6,8 @@
     "license" : "BSD-3-Clause",
     "require" : {
         "php"          : ">=5.5",
-        "ext-mbstring" : "*"
+        "ext-mbstring" : "*",
+        "sabre/xml"    : "~0.2"
```
evert: This needs an extra digit. 0.2.1 is guaranteed to be compatible, but 0.3 is not.

lib/Component.php
```diff
@@ -346,6 +348,55 @@ function jsonSerialize() {
...
+            } else {
+                $properties[] = $child;
+            }
+        }
+
+        $writer->startElement(strtolower($this->name));
+
+        if (!empty($properties)) {
+
+            $writer->startElement('properties');
```
evert: I'd really like it to be possible to write:

```xml
$writer->writeElement('properties', $properties);
```

But this may be a sabre/xml todo..

lib/Node.php
```diff
@@ -9,7 +11,11 @@
  * @author Evert Pot (http://evertpot.com/)
  * @license http://sabre.io/license/ Modified BSD License
  */
-abstract class Node implements \IteratorAggregate, \ArrayAccess, \Countable {
+abstract class Node
+    implements \IteratorAggregate,
+               \ArrayAccess,
+               \Countable,
+               \JsonSerializable {
```
evert: yay!

lib/Parser/XML.php
```diff
@@ -0,0 +1,275 @@
...
+                                $propertyType = 'text';
+
+                                foreach ($xmlProperty['value'] as $xmlRequestChild) {
+                                    $propertyValue[static::getTagName($xmlRequestChild['name'])]
+                                        = $xmlRequestChild['value'];
+                                }
+                                break;
+
+                            case 'freebusy':
+                                $propertyType = 'freebusy';
```
evert: no break intentional?

lib/Parser/XML/Xcal.rng
```diff
@@ -0,0 +1,1192 @@
+# RELAX NG Schema for iCalendar in XML
```
evert: Wondering if this should move outside of the lib/ tree... Perhaps good to have in a separate `schema` directory in the root? Are there any cons to that?

evert:
I did a read-over of the commit. I'm overall super happy with the results and excited to merge this :)

lib/Component.php
Hywan: This already exists: https://github.com/fruux/sabre-xml/blob/master/lib/Writer.php#L192-L200. Would you prefer this form?

lib/Parser/XML.php
Hywan: Yup.

lib/Parser/XML/Xcal.rng
Hywan: In Hoa, we always put all the resources inside the library, and we access it through `hoa://Library/<name>/<path>/`, for instance `hoa://Library/Ruler/Grammar.pp` to access to the grammar of `Hoa\Ruler`. This is very convenient.

But, according to the structure of sabre libraries, yup, maybe it's better to have a `resources/schema/` directory? I am going to create it.

lib/Component.php
evert: It would save 4 lines, so I think that'd be nice :)

lib/Parser/XML.php
evert: For 'cases' where there's no break, I think it's important to add an inline docblock that explains why.. or at the very least one that says 'no break intentional'.

Hywan: Ok.

Hywan: Done. Github does not hide this comment thread because the diff is just one line bellow, but it's done ;-).

lib/Component.php
Hywan: So, I cannot use `Xml::writeElement` here because we loop over an array, and `Xml::write`  supports only array of (`name`, `value`, `attributes`) tuples: It does not support an array of `XmlSerializable` objects. But I have update all the other parts of the code. See https://github.com/fruux/sabre-vobject/commit/6b6361088384b79b42fef16306e14abd2b4fb70f.

Hywan:
:gift:

Hywan:
\o/

PR #178, opened 2015-01-07
Skip the BOM if exists

Hywan:
Fix https://github.com/fruux/sabre-vobject/issues/175.
Fix https://github.com/owncloud/contacts/issues/635.

We skip the [BOM](http://graphemica.com/FEFF) if exists.

This PR addresses only `Parser/MimeDir.php` since I don't think it is needed in `Parser/Json.php` nor (the incoming) `Parser/Xml.php`.

### DIFF
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -97,6 +97,16 @@ class MimeDir extends Parser {
     protected function parseDocument() {
 
         $line = $this->readLine();
+
+        // BOM is ZERO WIDTH NO-BREAK SPACE (U+FEFF).
+        // It's 0xEF 0xBB 0xBF in UTF-8 hex.
+        if (   3 <= strlen($line)
+            && ord($line[0]) === 0xef
+            && ord($line[1]) === 0xbb
+            && ord($line[2]) === 0xbf) {
+            $line = substr($line, 3);
+        }
+
         switch(strtoupper($line)) {
             case 'BEGIN:VCALENDAR' :
                 $class = isset(VCalendar::$componentMap['VCALENDAR'])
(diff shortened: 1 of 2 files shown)


jbtbnl:
@dagbdagb it would be cool if you could check whether this fix solves https://github.com/owncloud/contacts/issues/635!

dagbdagb:
It sure would. 
But the version of vobject employed by OC is ancient. Not sure if current vobject is a drop-in replacement. Will see, but this is possibly a few steps beyond my capabilities to test *in OC*. Kudos to @Hywan .

Hywan:
@dagbdagb Maybe it can help if you use the [`patch` command](http://www.freebsd.org/cgi/man.cgi?query=patch) instead of `git merge`.

Hywan:
@jbtbnl It's not possible to add a test case in ownCloud and then see, in a second time, if this patch fixes this test case?

jbtbnl:
@Hywan owncloud/core first needs to be updated to SabreDAV 2.x and Sabre/VObject 3.x but that has been postponed till after the release of ownCloud 8: https://github.com/owncloud/core/issues/12876

evert:
I think this patch should be rewritten in a way that it doesn't rely on a stream to be seekable.

On many PHP versions (pre-5.6 I believe) `php://input` is not seekable, so `fseek()` will fail for pretty much any `PUT` request I believe.

evert:
This is also why the readLine() method is so complex, because it needs to look ahead from time to time, so it has a special 'next line buffer'.

But unless I'm mistaken, I think you could simply add the BOM check to after `$line = $this->readLine();` no?

Hywan:
You're right about `php://input` is not always seekable. Strangely, tests are green on PHP5.5 but http://3v4l.org/7kgfk tends to say the contrary.

I am going to update the PR.

Hywan:
Updated and rebased.

evert:
:+1:

PR #180, opened 2015-01-10
Validator upgrades.

### LINKED ISSUE #176: vobject claims UID MUST be present, does not automatically fix up missing UID
I note that vobject loudly complains if a vcard misses a UID:

vCard: 3.0
  [ERROR] UID MUST appear exactly once in a VCARD component

According to RFC 6350, UID has    Cardinality:  *1

...which means:
    |      *1     | Exactly one instance per vCard MAY be present.   |

Now, RFC 6350 covers vcard 4.0. And also claims to replace RFC 2426 (vcard 3.0).
RFC 2426 only lists FN, N and VERSION as mandatory. 
(side note: N is mandatory in vcard 3.0, but not 4.0).

In short:
I believe vobject is in error claiming a UID MUST be present per vcard.

However, *if* the project manager still thinks this is important to keep, how about autogenerating a UID for those vcards missing one?

evert:
This comment adds two new options to the validator:

1. PROFILE_CARDDAV
2. PROFILE_CALDAV

Passing these options will tell the validator to assume that the
vcard/icalendar object needs to be validated specifically according to
the rules of caldav/carddav.

For vCards specifically this means that the UID is required for CardDAV,
but we only emit a warning without PROFILE_CARDDAV.

For iCalendar there's a bunch of new rules that we will eventually
integrate into sabredav.

This fixes #176.

### DIFF
--- a/lib/Component.php
+++ b/lib/Component.php
@@ -47,7 +47,7 @@ class Component extends Node {
      * @param bool $defaults
      * @return void
      */
-    public function __construct(Document $root, $name, array $children = array(), $defaults = true) {
+    function __construct(Document $root, $name, array $children = array(), $defaults = true) {
 
         $this->name = strtoupper($name);
         $this->root = $root;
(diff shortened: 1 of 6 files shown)


Hywan:
:+1:

Hywan:
Is documentation up-to-date?

evert:
It's not. I'll update a separate ticket for this.

Hywan:
:+1:

Issue #94, opened 2014-06-03
Quote parameters with colons, semicolons or commas according to RFC 5545


thomascube:
As defined in [RFC 5545](https://tools.ietf.org/html/rfc5545#section-3.2), parameter values with colons, semicolons or commas should be quoted instead of escaping those characters.

I witnessed some Android sync clients having troubles with escaped instead of quoted property parameters. This was partially fixed in #65 but only for colons.

PR #81, opened 2014-01-23
Fix double-escaping of , and ; in compound properties

thomascube:
This is a fallow-up to my previous pull request https://github.com/fruux/sabre-vobject/pull/66 now with unit tests. It fixes the wrong double-escaping of compound properties with commas and semicolons.

PR #182, opened 2015-01-12
Fix double-escaping of , and ; in compound properties

thomascube:
This is a fresh version of my previous pull requests #66 and #81. It fixes the wrong double-escaping of compound properties with commas and semicolons for the 2.1 branch of vobject. This is a proposed fix for issue #94.

### DIFF
--- a/lib/Sabre/VObject/Property.php
+++ b/lib/Sabre/VObject/Property.php
@@ -195,6 +195,15 @@ class Property extends Node {
             '\n',
             '',
         );
+
+        // avoid double-escaping of \, and \; from Compound properties
+        if (method_exists($this, 'setParts')) {
+            $src[] = '\\\\,';
+            $out[] = '\\,';
+            $src[] = '\\\\;';
+            $out[] = '\\;';
+        }
+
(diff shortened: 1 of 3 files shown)


lib/Sabre/VObject/Property.php
```diff
@@ -195,6 +195,15 @@ public function serialize() {
             '\n',
             '',
         );
+
+        // avoid double-escaping of \, and \; from Compound properties
+        if (method_exists($this, 'setParts')) {
```
Hywan: Would not be better to filter by instance names instead of the method existence?

thomascube: The reason why I check by method is that the replacements below revert the escape characters that get added in `setParts()`. IMO the check is just right.

Hywan:
Should not we fix the first escape instead of reverting a previous escape?

evert:
@Hywan yes, we should... but that required a major BC break, so we gave it version 3.0 :)

evert:
So this patch attempts to partially fix the issue in a way that doesn't introduce a BC break in the 2.1 branch.

Hywan:
:-p

PR #188, opened 2015-01-21
Keep floating time on Recur\EventIterator::getEventObject

armin-hackmann:
to correctly expand with reference timezone (if any).

This fixes a problem where recurrent events with floating times would get set to UTC by EventIterator::getEventObject() while it should be transferred to UTC in VCalendar::expand().

### DIFF
--- a/tests/VObject/Recur/EventIterator/ExpandFloatingTimesTest.php
+++ b/tests/VObject/Recur/EventIterator/ExpandFloatingTimesTest.php
@@ -1,0 +1,122 @@
+<?php
+
+namespace Sabre\VObject;
+
+use
+    DateTime,
+    DateTimeZone;
+
+/**
+ * This is a unittest for Issue #53.
+ */
+class ExpandFloatingTimesTest extends \PHPUnit_Framework_TestCase {
+
+    function testExpand() {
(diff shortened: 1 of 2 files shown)


Hywan:
:+1:

evert:
Nice!

tests/VObject/Recur/EventIterator/ExpandFloatingTimesTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+namespace Sabre\VObject;
+
+use
+    DateTime,
+    DateTimeZone;
+
+/**
+ * This is a unittest for Issue #53.
+ */
```
DominikTo: No :-) 53 is about something else, I think.

evert: You're right

armin-hackmann: Argh, my bad! Didn't recheck after duplicating the file from another test. :/

Hywan:
The Gang of Bad Reviewers™

PR #190, opened 2015-01-21
Add the collection object: A set of components

Hywan:
Related to https://github.com/fruux/sabre-vobject/issues/189.
It does not fix this issue but this is needed to solve it.

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -1,0 +1,132 @@
+<?php
+
+namespace Sabre\VObject;
+
+/**
+ * Collection
+ *
+ * A collection represents a set of documents. See Sabre\VObject\Document. It's
+ * not a bag, thus all items must have the same type.
+ *
+ * @copyright Copyright (C) 2011-2015 fruux GmbH (https://fruux.com/).
+ * @author Ivan Enderlin
+ * @license http://sabre.io/license/ Modified BSD License
+ */
+class Collection implements \ArrayAccess, \Countable, \IteratorAggregate {
+
+    /**
+     * Type of document.
+     *
+     * @var string
+     */
+    protected $type;
+
+    /**
+     * Collection.
+     *
+     * @var array
+     */
+    protected $collection = [];
+
+    /**
+     * Construct the collection by setting the type of items it contains, for
+     * instance: 'Sabre\VObject\Vcard'.
+     *
+     * @param string $type    Type of items.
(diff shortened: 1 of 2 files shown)


lib/Collection.php
```diff
@@ -0,0 +1,132 @@
...
+     * @param  mixed  $offset    Component offset.
+     * @param  mixed  $value     Component.
+     * @return Collection
+     * @throw  \InvalidArgumentException
+     */
+    function offsetSet($offset, $value) {
+
+        $type = $this->type;
+
+        if (!(($value instanceof Document) && ($value instanceof $type))) {
```
staabm: couldn't you instead check whether `$type` is `is_subclass_of Document` in `__construct` and check here only check `instanceof $type` then?

Hywan: Unfortunately, we cannot have `$type instanceof Document` where `$type` is a string.

Hywan:
The build fails on CS that are not related to this PR. These CS are, by the way, fixed in #160.

lib/Collection.php
staabm: therefore I suggest `is_subclass_of`

Hywan: :+1:

Hywan:
I wonder if we need to define `Collection` as a `Node`. It would help a lot. I met this usecase while developing xCard support. I am going to update this PR.

Hywan:
This PR **must not** be merged before #160, because `Node` changes.

Hywan:
Actually, extending `Node` will not help. We have to extend `Component` in order to work seamlessly with `Writer`. Hmm… I think it's better to keep `Collection` as it is and extend the `Writer` to deal with it.

evert:
I see this is already closed, but I would wonder.. is this needed at all? The splitter classes exists because it allows someone to progressively parse many objects, without loading them all in memory. It looks like the intent of this class is to actually load everything in memory.

I feel that for many real-life use-cases, this would not be desirable.

One idea could be to merge the 'getNext()` method from the splitter right into the parser. That way you can start parsing ones, and if you know there may be more in a file, you can call it multiple times.

Regardless, since we can start depending on PHP 5.5, whenever you consider building an iterator, consider using a generator instead ;)

Hywan:
@evert That was exactly my plan :-]. You read in my minds!

PR #195, opened 2015-01-22
Fixed short array syntax for PHP < 5.4

rickdenhaan:
Fixes https://github.com/fruux/sabre-vobject/issues/194


evert:
Well that makes sense.. These two files are indeed outside of what's being tested in php unit, and largely 'utility scripts' :)

Thanks for the patch!

PR #196, opened 2015-01-23
Making the recurrence iterator faster on large calendars.

evert:
This is in response to a 5MB calendar I got from @armin-hackmann, with many, many `RRULE's` and overrides.

I found out that a major slowdown actually came from traversing the list of events in VCALENDAR just to find all the events with a particular `UID`.

I fixed it by doing the lookup once for the entire calendar, rather than letting the recurrence iterator do it (which does it once for every event that needs expanding).

The speed-up was quite huge.

```
Before:

__global__  |||||||||||||||||||||||||||||||||||||||||||||||||||| 289336ms, 100.0%
parse       ||                                                   13353ms,   4.6%
expand      ||||||||||||||||||||||||||||||||||||||||||||||||||   275983ms,  95.4%

After

__global__  |||||||||||||||||||||||||||||||||||||||||||||||||||| 28022ms, 100.0%
parse       |||||||||||||||||||||||||                            13632ms,  48.6%
expand      |||||||||||||||||||||||||||                          14389ms,  51.4%
```

This is a 1900% increase in speed ;) 

One other thing this tells me, is that grabbing values from properties is _super slow_. This is because when you do something like this:

```php
echo $vevent->UID->getValue();
```

vobject will actually iterate over _every_ property defined on vevent, and see if the name matches UID.
I think we can speed this up quite a bit by indexing all properties in an array by their property name.

But this will break the `$children` array, which is currently protected, so this needs to be a 4.x change.

### DIFF
--- a/bin/rrulebench.php
+++ b/bin/rrulebench.php
@@ -1,0 +1,32 @@
+<?php
+
+include __DIR__ . '/../vendor/autoload.php';
+
+if ($argc < 4) {
+    echo "sabre/vobject ", Sabre\VObject\Version::VERSION, " RRULE benchmark\n";
+    echo "\n";
+    echo "This script can be used to measure the speed of the 'recurrence expansion'\n";
+    echo "system.";
+    echo "\n";
+    echo "Usage: " . $argv[0] . " inputfile.ics startdate enddate\n";
+    die();
+}
+
+list(, $inputFile, $startDate, $endDate) = $argv;
+
+$bench = new Hoa\Bench\Bench();
+$bench->parse->start();
+
+echo "Parsing.\n";
+$vobj = Sabre\VObject\Reader::read(fopen($inputFile,'r'));
+
+$bench->parse->stop();
+
+echo "Expanding.\n";
+$bench->expand->start();
+
+$vobj->expand(new DateTime($startDate), new DateTime($endDate));
+
+$bench->expand->stop();
+
+echo $bench,"\n";
--- a/lib/Component/VCalendar.php
+++ b/lib/Component/VCalendar.php
@@ -35,12 +35,14 @@ class VCalendar extends VObject\Document {
      * @var array
      */
     static $componentMap = array(
-        'VALARM'    => 'Sabre\\VObject\\Component\\VAlarm',
-        'VEVENT'    => 'Sabre\\VObject\\Component\\VEvent',
-        'VFREEBUSY' => 'Sabre\\VObject\\Component\\VFreeBusy',
-        'VJOURNAL'  => 'Sabre\\VObject\\Component\\VJournal',
-        'VTIMEZONE' => 'Sabre\\VObject\\Component\\VTimeZone',
-        'VTODO'     => 'Sabre\\VObject\\Component\\VTodo',
+        'VALARM'        => 'Sabre\\VObject\\Component\\VAlarm',
+        'VEVENT'        => 'Sabre\\VObject\\Component\\VEvent',
+        'VFREEBUSY'     => 'Sabre\\VObject\\Component\\VFreeBusy',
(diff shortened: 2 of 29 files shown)


lib/Recur/EventIterator.php
```diff
@@ -73,38 +73,44 @@ class EventIterator implements \Iterator {
...
+     *
+     * The $uid parameter is only required for the first method.
+     *
+     * @param Component|array $input
      * @param string|null $uid
      * @param DateTimeZone $timeZone Reference timezone for floating dates and
      *                               times.
      */
-    public function __construct(Component $vcal, $uid = null, DateTimeZone $timeZone = null) {
+    public function __construct($input, $uid = null, DateTimeZone $timeZone = null) {
```
staabm: isnt this a BC break, or might at least trigger E_STRICT warnings in subclasses of user-apps?

lib/Component/VCalendar.php
```diff
@@ -486,5 +494,26 @@ function validate($options = 0) {
...
+    /**
+     * Returns all components with a specific UID value.
+     *
+     * @return array
+     */
+    function getByUID($uid) {
+
+        return array_filter($this->children, function($item) use ($uid) {
 
+            if (!$item instanceof Component) {
```
staabm: can all known components have a `UID` or can we filter more agressive based on the class here?

bin/rrulebench.php
```diff
@@ -0,0 +1,32 @@
...
+    echo "This script can be used to measure the speed of the 'recurrence expansion'\n";
+    echo "system.";
+    echo "\n";
+    echo "Usage: " . $argv[0] . " inputfile.ics startdate enddate\n";
+    die();
+}
+
+list(, $inputFile, $startDate, $endDate) = $argv;
+
+$bench = new Hoa\Bench\Bench();
```
Hywan: You need to add this as a requirement in `composer.json`.

Hywan:
Instead of going through all the children to find the `UID` entry, why not transforming `$children` into an hashmap: `children` × `child`? The access is going to be very fast.

bin/rrulebench.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
lyrixx: Missing `#!/usr/bin/env php`
and the file is not executable.

lyrixx:
I did that for you: https://blackfire.io/profiles/80fceb09-6ad4-4cd1-8665-93cc0bc10a04/graph

I used this [ics](https://www.google.com/calendar/ical/amnhastro@gmail.com/public/basic.ics) between 2001 and 2015.

So basically, I just ran `blackfire run --sample 20 php bin/rrulebench.php file:///home/greg/Downloads/basic.ics  2001 2015` after a fresh `git clone` + `composer install` ;)

Hywan:
@lyrixx Product placement spotted!

lib/Recur/EventIterator.php
evert: Constructor signature is not enforced when subclassing :) Should not be an issue.

lib/Component/VCalendar.php
evert: The only component that doesn't, is the VTIMEZONE.

bin/rrulebench.php
evert: I had that in an earlier commit as `require-dev`, but it would break the build as we still support PHP 5.3. I can add it back in on the master branch, right after a merge.

evert:
Really nice looking @lyrixx :). Product placement was indeed quite blatant though :P

bin/rrulebench.php
Hywan: Ok.

PR #201, opened 2015-01-30
LAST-MODIFICATION does not exist: It's LAST-MODIFIED.

Hywan:
According to [RFC5545](https://tools.ietf.org/html/rfc5545#section-3.8.7.3), there is no `LAST-MODIFICATION` property, only `LAST-MODIFIED`. This patch fixes this bug. No test is added.


evert:
This was fixed in fe021e2, I'm gonna check if that was merged to master at all.

evert:
I merged it now.. looks like there were a few lingering commits in 3.3. Sorry about this :(

Hywan:
@evert No it's fine ;-).

PR #204, opened 2015-02-05
milliseconds in date time parser

armin-hackmann:
We came across some cards (generated by an export on mailbox.org) that had milliseconds in a property-value (REV in this instance), for example:
```
BEGIN:VCARD
VERSION:3.0
PRODID:OPEN-XCHANGE
FN:Armin Hackmann
N:Hackmann;Armin;;;
X-OPEN-XCHANGE-CTYPE:contact
REV:20140905T151124.056Z
UID:d91b949e-69ef-41a8-a1c3-ff6a7743973c
END:VCARD
```

If we'd want to convert a card containing this to another version, the DateTimeParser would exit with an InvalidArgumentException. While the behavior is correct (according to the spec, there shouldn't be a property-value containing milliseconds), I wonder if we should support parsing such a date-time/time value?

### DIFF
--- a/lib/DateTimeParser.php
+++ b/lib/DateTimeParser.php
@@ -199,7 +199,7 @@ class DateTimeParser {
      * This method returns an array, not a DateTime value.
      *
      * The elements in the array are in the following order:
-     * year, month, date, hour, minute, second, millisecond, timezone
+     * year, month, date, hour, minute, second, timezone
      *
      * Almost any part of the string may be omitted. It's for example legal to
      * just specify seconds, leave out the year, etc.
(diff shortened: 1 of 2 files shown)


Hywan:
You're right, it's not valid regarding any RFC: iCalendar, vCard, jCal, jCard, xCal and xCard. None of them support milliseconds :-/. However, to avoid an error, we should parse and drop it (parse but not dump the `millisecond` key in the array computed by the datetime parsers). So basically, no `?P<milisecond>`.

Thoughts?

evert:
I guess we could start ignoring those values... Did you guys try to report a bug at mailbox.org?

evert:
I didn't see the full thread here, but just wanted to point out that I agree with @Hywan . The right approach is to discard the value, not keep it.

armin-hackmann:
Updated & the value now gets discarded.

Issue #126, opened 2014-09-16
Attendees updating their status for 1 instance of a recurring event is triggered as a 'significant change'


evert:
##

armin-hackmann:
I digged a bit into this and while I'm not 100% certain that this is the same error, I could replicate a problem when a recurring event with exdates gets insignificantly updated (e.g. a new alarm via iCal) and the client changes the order of `EXDATE` in the ics.

You can find a test in my fork:
https://github.com/armin-hackmann/sabre-vobject/commit/4fb6c06a8f1422dd001495095e73f24bc5096284

The [ITip Broker](https://github.com/fruux/sabre-vobject/blob/3.3/lib/ITip/Broker.php#L917) simply chains all properties and their values marked to significantly change an event so its pretty obvious why a change is assumed.

We could sort the `EXDATE` values before adding it ([see this commit](https://github.com/armin-hackmann/sabre-vobject/commit/3a77fb067999c28dbb1d0c44d982f791005ea59a)). I'm not certain if this is the best way to go about it thus 'just' this comment and not a PR. Did I miss something?

alecpl:
Another case. Server has `RRULE:FREQ=WEEKLY;INTERVAL=1`, client sends `RRULE:FREQ=WEEKLY`.

PR #205, opened 2015-02-05
Dealing with EXDATES in ITip\Broker & fix for #126

armin-hackmann:
This PR solves two issues:
* parseEventInfo() only returned the value of the first `EXDATE` present in an object
* if an application swapped `EXDATE`-values in a PUT, the `$significantChangeHash` would imply a real change of the event happened

Also see the initial [issue](https://github.com/fruux/sabre-vobject/issues/126) & the thread in [this commit](https://github.com/armin-hackmann/sabre-vobject/commit/3a77fb067999c28dbb1d0c44d982f791005ea59a).


lib/ITip/Broker.php
```diff
@@ -855,7 +855,10 @@ protected function parseEventInfo(VCalendar $calendar = null) {
                 $sequence = $vevent->SEQUENCE->getValue();
             }
             if (isset($vevent->EXDATE)) {
-                $exdate = $vevent->EXDATE->getParts();
+                foreach ($vevent->select('EXDATE') as $val) {
+                    $exdate = array_merge($exdate, $val->getParts());
```
Hywan: Do we consider the `PRIORITY` property here to sort `EXDATE` before merging them? Because order is important when merging.

armin-hackmann: As discussed via Slack yesterday you: there is no relation between PRIORITY and EXDATE. And the exdates get sorted (so we have earliest to latest in our variable) after merging values of all EXDATE-properties.

Did I miss something?

evert: No, you are good :+1:

armin-hackmann:
anything missing for you, @evert? :)

evert:
Sorry for the delay armin. Just noticed this on my todo list from a few days ago :S

PR #215, opened 2015-03-26
Test for outlook encoding issue

DominikTo:
Outlook encodes characters (like Umlauts) like this:

```
FN;ENCODING=QUOTED-PRINTABLE:Test=DF=E4=F6=FC=C4=D6
```

When we convert to vCard 4.0 and then to jCard it seems like the conversion fails, as we're not converting to the expected result: `TestßäöüÄÖÜ`

### DIFF
--- a/tests/VObject/VCardConverterTest.php
+++ b/tests/VObject/VCardConverterTest.php
@@ -308,7 +308,7 @@ IN;
         $vcard = Reader::read($input);
         $vcard = $vcard->convert(Document::VCARD40);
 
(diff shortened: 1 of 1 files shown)


evert:
If the fruux outlook plugin does this, it should switch to UTF-8. Still a valid ticket though.

DominikTo:
No, fruux plugin is not affected. Was reported by a third party plugin.

DominikTo:
Closing this one in favor of #255.

Issue #217, opened 2015-04-02
VCard\Date incorrectly serialized when initialized with DateTime object


thomascube:
Consider the following test case:
```php
$datetime = new \DateTime('2014-04-02');
$vcard = new VObject\Component\VCard();
$prop = $vcard->createProperty('BDAY', $datetime, null, 'DATE');
echo $prop->serialize();
````
Results in
```
BDAY:20140402T000000Z
```
instead of the expected
```
BDAY;VALUE=DATE:2014-04-02
```

thomascube:
PR will follow soon...

thomascube:
Sorry, the correct result would be either
 `BDAY:20140402` according to RFC 6350
or
 `BDAY:2014-04-02` according to RFC 2426.

evert:
They are actually both valid. I forget which we prefer. I would personally prefer whatever 'most people' are doing, but not sure which one of those it is.

evert:
Go for the format with dashes =)

thomascube:
The DateAndOrTime class uses Ymd\\THis for date formatting, that's RFC 6350 as far as I can tell.

evert:
Hm yes it does say that the extended format is now allowed. Sounds good then!

PR #218, opened 2015-04-02
Fix Property\VCard\Date to serialize DateTime objects as date-only string

thomascube:
Fixes issue #217


evert:
Great patch, thank you!

PR #224, opened 2015-05-07
Saving some memory for echo.

dseguy:


Hywan:
Maybe a rebase from master would be a good idea ;-).

evert:
Yea, the suggestion is fine but it needs to either target the master branch, or the patch needs to be creaed _for_ the 3.4 branch. Right now it was written for the master branch, but the 3.4 branch is targetted, resulting in the 250+ commits.

dseguy:
Let me do this on the 3.4 branch. I'll push a new PR then.

PR #230, opened 2015-06-09
vEvent->isInTimeRange() fails for all-day events that end on start of given time range

armin-hackmann:
`DTEND` should be non inclusive (see http://tools.ietf.org/html/rfc5545#page-54), but `VEVENT->isInTimeRange()` will return events that ended the day before. Additionally the timezone of the timerange should be considered.

Unit tests for these two cases are included as well as a fix for each.

### DIFF
--- a/lib/Component/VEvent.php
+++ b/lib/Component/VEvent.php
@@ -53,6 +53,9 @@ class VEvent extends VObject\Component {
             // See:
             // http://tools.ietf.org/html/rfc5545#page-54
             $effectiveEnd = $this->DTEND->getDateTime();
+            if (!$this->DTEND->hasTime()) {
+                $effectiveEnd->modify('-1 second');
+            }
 
         } elseif (isset($this->DURATION)) {
             $effectiveEnd = clone $effectiveStart;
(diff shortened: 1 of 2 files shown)


lib/Component/VEvent.php
```diff
@@ -60,6 +63,7 @@ public function isInTimeRange(\DateTime $start, \DateTime $end) {
         } elseif (!$this->DTSTART->hasTime()) {
             $effectiveEnd = clone $effectiveStart;
             $effectiveEnd->modify('+1 day');
+            $effectiveEnd->modify('-1 second');
```
Hywan: Could we use a modifier like `last second of this day` or `front of tomorrow`? Not tested but interesting though.

lib/Component/VEvent.php
```diff
@@ -52,7 +52,10 @@ public function isInTimeRange(\DateTime $start, \DateTime $end) {
             //
             // See:
             // http://tools.ietf.org/html/rfc5545#page-54
-            $effectiveEnd = $this->DTEND->getDateTime();
+            $effectiveEnd = $this->DTEND->getDateTime($end->getTimezone());
+            if (!$this->DTEND->hasTime()) {
+                $effectiveEnd->modify('-1 second');
```
evert: Instead of doing `-1 second`, I think you can just alter the condition in the bottom of the file from a `<=` into a `<`, no?

DominikTo: Removing this and the other "-1" along with just doing a `<` comparison would at least also pass the existing tests, but not 100% sure if it might be wrong for any of the other cases.

evert: If the tests pass, I'm reasonably confident that it's ok.

DominikTo:
New PR in #231.

PR #231, opened 2015-06-09
vEvent->isInTimeRange() fails for all-day events that end on start of given time range 

DominikTo:
New PR for #230.


armin-hackmann:
Looks good to me :+1:

PR #233, opened 2015-06-23
Identify "+00:00" as UTC

c960657:
When a DateTime object is created from a Unix timestamp using the @ syntax, `new \DateTime('@123456789')`, the resulting object gets a timezone with the name `+00:00` (tested with PHP 5.5.25 and 5.4.38). Passing a timezone object as the second argument does not change this.

This timezone represents UTC, but it is not identified as a UTC timezone in VObject\Property\ICalendar\DateTime.

This PR adds `+00:00` to the list of known aliases for UTC.

(BTW, there are [other aliases](http://php.net/manual/en/timezones.others.php), but they are all deprecated)

### DIFF
--- a/lib/Property/ICalendar/DateTime.php
+++ b/lib/Property/ICalendar/DateTime.php
@@ -224,7 +224,7 @@ class DateTime extends Property {
                 }
                 if (is_null($tz)) {
                     $tz = $d->getTimeZone();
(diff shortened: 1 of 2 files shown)


evert:
Thanks for the bug report. Travis failure is unrelated. Going to fix that on master first.

evert:
Could you rebase/merge from master so we get a correct message from travis?

c960657:
Done :-)

PR #234, opened 2015-06-24
Add @property for dynamic properties to phpdoc

Henni:
IDEs don't know about dynamic properties accessed via __get() and __set(). To fix this @property can be used in the phpdoc to indicate which properties are available.

Similar changes should probably be also applied to other components like VCard.

### DIFF
--- a/lib/Component/VCalendar.php
+++ b/lib/Component/VCalendar.php
@@ -13,6 +13,59 @@ use Sabre\VObject\Recur\NoInstancesException;
  * The VCalendar component.
  *
  * This component adds functionality to a component, specific for a VCALENDAR.
+ * 
+ * @property \Sabre\VObject\Property\FlatText $CALSCALE
+ * @property \Sabre\VObject\Property\FlatText $METHOD
+ * @property \Sabre\VObject\Property\FlatText $PRODID
+ * @property \Sabre\VObject\Property\FlatText $VERSION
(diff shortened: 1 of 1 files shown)


staabm:
@Henni did you manually create those properties or have you generated those using a small snippet? it might make sense to automate such things, because they get very easily out of sync with the code.

Henni:
I did it manually.
It may be difficult to automate this as not every component lists its properties inside a propertyMap as vcalendar and vcard do.

Is the Travis failure normal?

evert:
The travis failure is happening because there's a space somewhere at the end of a line.

I'm not super sure about this patch though. It's not really a good list, as different components have different properties, and this is just a massive list of properties, all assigned to the VCALENDAR object, many of which don't actually belong there.

It also doesn't address extensions well. If there was some dynamic way to tell IDE's it would be a lot better, but I'm just not sure about this big list, as it's going to be hard to keep accurate...

staabm:
> It also doesn't address extensions well. If there was some dynamic way to tell IDE's it would be a lot better, but I'm just not sure about this big list, as it's going to be hard to keep accurate...

that was also my impression. such thing would only be worthwhile when we could automate the creation (or find a generic way which is IDE supported as you pointed out)

evert:
No further feedback received, going to close this for the time being.

PR #239, opened 2015-07-07
Add BirthdayCalendarGenerator

DominikTo:


tests/VObject/VCardBirthdayTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+class VCardBirthdayTest extends TestCase {
+
+    function testValidBirthday() {
+
+        $input = <<<VCF
+BEGIN:VCARD
+VERSION:3.0
+N:Gump;Forrest;;Mr.
+FN:Forrest Gump
```
Hywan: :yum:

Hywan:
:+1: for me. Good job!

DominikTo:
I wonder:

1. Should `$format` support using first name and last name separately (e.g. if a `SUMMARY` like "John's Birthday" (instead of "John Doe's Birthday") is desired?
2. Should this method support falling back to `N`?
3. Should this method maybe return null for vCards that have the proprietary `X-ABShowAs:COMPANY` set?
4. Should we optionally generate multiple recurring instances, so we could support `SUMMARY`s like "John Doe's 30. Birthday"?

evert:
The only concern I have, is that I feel it might be better for this functionality to live outside of the vcard component. It's nice to keep the actual method that's in there, but perhaps there should be a separate BirthdayCalendarGenerator class.

Two reasons:

1. The functionality is very similar to FreebusyGenerator. Ideally they also have very similar API's
2. I imagine people will want to create birthdays calendars for multiple vcards at the same time. This would only allow creating 1 birthday VCALENDAR, so this would force people to run this multiple times and then merge afterwards. If there's an object similar to FreebusyGenerator, you could feed it multiple vcards.

tests/VObject/VCardBirthdayTest.php
```diff
@@ -0,0 +1,120 @@
...
+BDAY:19850407
```
evert: Hey that's my birthday! Are you calling me forest gump? :P

Regardless... you need to add a test for BDAY values without a year and without a month.

evert:
It would also be real nice if you support `ANNIVERSARY` =)

tests/VObject/VCardBirthdayTest.php
DominikTo: Haha, no, I just added an easter egg and with your name it would have been too obvious. ;-)

DominikTo:
Good idea with the BirthdayCalendarGenerator, now I also remember that we talked about this. =)

Hywan:
`BirthdayCalendarGenerator` is better.

lib/BirthdayCalendarGenerator.php
```diff
@@ -0,0 +1,204 @@
...
+     */
+    protected $timeZone;
+
+    /**
+     * Default year.
+     * Used for dates without a year.
+     *
+     * @var int
+     */
+    protected $defaultYear = 1900;
```
DominikTo: Maybe, once #242 is merged, the year defined in `$minDate` in the new `Settings` class should be used instead.

lib/BirthdayCalendarGenerator.php
```diff
@@ -0,0 +1,204 @@
...
+            } catch (\InvalidArgumentException $e) {
+                continue;
+            }
+
+            // Set a year if it's not set.
+            if (!$dateParts['year']) {
+                $object->BDAY=$this->defaultYear . '-' . $dateParts['month'] . '-' . $dateParts['date'];
+
+                // @TODO:
+                // It may make sense to add an info to our event that the year was unknown.
```
DominikTo: Not sure about this, but maybe it's helpful for a client to know about this (e.g. to make sure that no "age" gets calculated for such events).

lib/BirthdayCalendarGenerator.php
evert: I don't know if I came up with 1900 here... but I would actually suggest putting this a lot further in the future. Putting it on 1900 means that usually 115 recurrence instances will be created to calculate birthdays, and this will not be very efficient. I'd set it to 2010 I think.

lib/BirthdayCalendarGenerator.php
evert: I agree, this does make sense.... perhaps we should add it as another value to X-SABRE-BIRTHDAY.

DominikTo:
Ready for review.

evert:
* All events need to have `VALUE=DATE`.
* Timezone is not relevant for those, so you can strip all timezone-related stuff.

lib/BirthdayCalendarGenerator.php
```diff
@@ -0,0 +1,212 @@
...
+     */
+    protected $timeZone;
+
+    /**
+     * Default year.
+     * Used for dates without a year.
+     *
+     * @var int
+     */
+    protected $defaultYear = 2000;
```
evert: Should probably be a constant.

lib/BirthdayCalendarGenerator.php
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        $this->objects = [];
+        foreach ($objects as $object) {
+
+            if (is_string($object)) {
+
+                $vObj = Reader::read($object);
+                if (!$vObj instanceof Component\VCard) {
+                    throw new \InvalidArgumentException('String could not be parsed as \\Sabre\\VObject\\Component\\VCard by setObjects');
```
evert: This case is not tested.

lib/BirthdayCalendarGenerator.php
```diff
@@ -0,0 +1,212 @@
...
+
+                $this->objects[] = $vObj;
+
+            } elseif ($object instanceof Component\VCard) {
+
+                $this->objects[] = $object;
+
+            } else {
+
+                throw new \InvalidArgumentException('You can only pass strings or \\Sabre\\VObject\\Component\\VCard arguments to setObjects');
```
evert: This case is not tested.

DominikTo:
Thanks!

* All events now have VALUE=DATE.
* Timezone related stuff is stripped.
* $defaultYear is now a DEFAULT_YEAR constant.
* Added tests for previously untested stuff.

DominikTo:
Looks like this can't get merged automatically though. Should I merge master into this branch or rebase on master?

evert:
Merge master into this branch =)

DominikTo:
Done.

evert:
Awesome!

Hywan:
Awesome \o/.

CHANGELOG.md
```diff
@@ -37,8 +37,12 @@ ChangeLog
 * #197: The `$children` property on components has been changed from `public`
   to `protected`. Use the `children()` method instead to get a flat list of
   objects.
+<<<<<<< HEAD
```
Hywan: Haaa damned, it is not supposed to be here :wink:.

Hywan: Addressed in https://github.com/fruux/sabre-vobject/commit/87c6a50dd93506784ac70ac228e62ef93f0db262. Good.

PR #241, opened 2015-07-08
Fixed a few typos

DominikTo:


lib/Component/VCard.php
```diff
@@ -357,7 +357,7 @@ function getValidationRules() {
 
             // FN is commented out, because it's already handled by the
             // validate function, which may also try to repair it.
-            // 'FN'           => '+',
+            // 'FN'           => '1',
```
evert: Are you sure about this?

evert: Nope it's wrong. http://tools.ietf.org/html/rfc6350#section-6.2.1 says:

    Cardinality:  1*

Which means 'at least one, but can be more.'

DominikTo: Haha, isn't it kinda late (or early)? 😴

The RFC [says](https://tools.ietf.org/html/rfc6350#section-6.2.1) *1 ("One **or more** instances per vCard MUST be present"), so according to the RFC "+" is correct, but it didn't make sense to me. Why would there ever be more than one FN?

Aren't we also enforcing just 1 FN here?
https://github.com/fruux/sabre-vobject/blob/master/lib/Component/VCard.php#L269

Happy to revert this one though =) Let me know.

evert: If I recall correctly, one use-case for multiple `FN` properties is different language variations. There's a `LANG` property as well for that purpose.

evert: And you are right that validator function is wrong.

DominikTo: Ah, that makes sense then. Rolling back this one.

PR #244, opened 2015-07-08
PHP 7 and HHVM  support

### LINKED ISSUE #243: Make sure we're passing PHP7 tests
Currently we don't support PHP7, because we're using a few class names that are now proprietary. (String, Float, Int). We'll need to rename these in order to support PHP7.

evert:
In this PR i'll make sure that both HHVM and PHP7 are supported. HHVM was supported before, but broken as of a little while due to a php incompatibility.

PHP7 no longer allows keywords such as Integer and Float to be used as classnames, so we're renaming those.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ matrix:
   fast_finish: true
   allow_failures:
     - php: hhvm
-    - php: 7
 
 script:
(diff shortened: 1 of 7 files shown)


Hywan:
:+1:

PR #247, opened 2015-07-23
Parser/MimeDir issue and timezone issue

DominikTo:
This test reproduces two issues:

1)
The `DESCRIPTION` makes the parser fail. Seems like the line is not properly matched.
Parser fails with `Sabre\VObject\ParseException: Invalid Mimedir file. Line starting at 28 did not follow iCalendar/vCard conventions`. When the `DESCRIPTION` property is removed, the event is parsed without issues.

```
(
    [name] =>
    [parameters] => Array
        (
        )

    [value] =>
)
```

2)
Running the test without the `DESCRIPTION` property, the test still fails, because we cannot properly identify the timezone: `InvalidArgumentException: We were unable to determine the correct PHP timezone for tzid: US-Central`.

### DIFF
--- a/tests/VObject/TimeZoneUtilTest.php
+++ b/tests/VObject/TimeZoneUtilTest.php
@@ -366,4 +366,50 @@ HI;
 
     }
 
+    function testEventFromSouthendzone() {
+
+        // from http://www.southendzone.com/ical/vikings.ics
+
+$vobj = <<<HI
+BEGIN:VCALENDAR
+VERSION:2.0
+X-WR-CALNAME:Minnesota Vikings
+X-WR-CALDESC:Scott Crevier, smc@smcnet.com
+PRODID:-//Scott Crevier//SouthEndZone.com//EN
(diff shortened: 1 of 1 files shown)


gjn:
TZOFFSETTO and TZOFFSETFROM might be used to determine correct timezone. Those properties are present in the VTIMEZONE sections under the STANDARD and DAYLIGHT sections.

evert:
The parser should fail because of the literal newlines.

DominikTo:
So should this test then get modified to only test for the timezone issue?

evert:
@DominikTo I'm opening a new ticket for this. The only real solution to this is what @gjn, and that's to start automatically detecting the timezone based on the `VTIMEZONE` values. I thought we have a ticket for it, but apparently we don't.

The test is not that useful because implementing that would likely require a lot more tests =)

evert:
Closing this ticket in favor of #248

PR #249, opened 2015-07-24
Workaround for owncloud/calendar issue #857

migrax:
Catch the NoInstancesException and return false as the result of
the isInTimeRange test.

I have tested this with a Jolla phone that also exhibits the error shown in issue #857.

### DIFF
--- a/lib/Component/VEvent.php
+++ b/lib/Component/VEvent.php
@@ -4,6 +4,7 @@ namespace Sabre\VObject\Component;
 
 use Sabre\VObject;
 use Sabre\VObject\Recur\EventIterator;
+use Sabre\VObject\Recur\NoInstancesException;
 
 /**
  * VEvent component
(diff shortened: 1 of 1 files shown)


evert:
Confirmed. Needs unittest before it can be merged.

DominikTo:
This has been dealt with in #253.

thinking-aloud:
since OC 8.0 I have to apply this fix after every little update. When will it get merged into the stable branch?

DominikTo:
This fix has been released in sabre/vobject v3.4.7 (https://github.com/fruux/sabre-vobject/commit/7b3690545451159fc608dfac6696110f6770ff5d) on 2015-09-04.

thinking-aloud:
After every update i get the old files and my phone won't sync. Even in the master branch, it is not as it should be. https://github.com/owncloud/3rdparty/blob/master/sabre/vobject/lib/Component/VEvent.php#L37

DominikTo:
The version you linked to, _is_ the fixed version from sabre/vobject 3.4.7 - which is what ownCloud is using (https://github.com/owncloud/3rdparty/blob/master/sabre/vobject/ChangeLog.md).

thinking-aloud:
> ownCloud 8.2.1 will ship with the updated sabre.
thank you, that sounds good

PR #250, opened 2015-07-29
Fix for isInTimeRange for recurring all day events in timezone

armin-hackmann:
Recurring all-day events are incorrectly included in time range requests when not using UTC in the time range.

e.g. (also see added unit test)
* All day event on every 1st of the month
* `isInTimeRange()` for the 2nd of a month 'Europe\Berlin`
* event would pass the test

This PR fixes that issue and a typo on initialization of `\Recur\EventIterator()`.


DominikTo:
:+1:

evert:
Awesome!

PR #251, opened 2015-07-30
Skip cards without FN in BirthdayCalendarGenerator

DominikTo:
Instead of failing on `FN->getValue()` we're now skipping cards that don't have a FN.

### DIFF
--- a/lib/BirthdayCalendarGenerator.php
+++ b/lib/BirthdayCalendarGenerator.php
@@ -124,6 +124,11 @@ class BirthdayCalendarGenerator {
             // VCardConverter handling the X-APPLE-OMIT-YEAR property for us.
             $object = $object->convert(Document::VCARD40);
 
+            // Skip if the card has no FN property.
+            if (!isset($object->FN)) {
+                continue;
+            }
+
             // Skip if the BDAY property is not of the right type.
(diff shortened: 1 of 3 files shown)


lib/BirthdayCalendarGenerator.php
```diff
@@ -136,6 +136,11 @@ function getResult() {
                 continue;
             }
 
+            // Skip if the card has no FN property.
+            if (!isset($object->FN)) {
```
staabm: I guess we should re-order this "skip if" statements from the cheapest to the expensive ones

DominikTo: Agree. Updated the PR.

DominikTo:
We _could_ also attempt to `->repair()` a vCard if there's no `FN`. The `repair()` method would then attempt to add a valid `FN` based on the `N` property.

Not sure if we should though.

DominikTo:
Here's an alternative that would attempt to fix a card without FN first: 770b54cb5cf50a7d226f7002efc49da0c6a700d5

evert:
I prefer this approach =)

DominikTo:
=)

Would it make sense to add an option that allows repairing?

evert:
No, it's not the responsibility of this class to do this.

PR #252, opened 2015-08-26
Empty BDAY properties break the BirthdayCalendarGenerator

DominikTo:


evert:
Are empty BDAY properties valid?

DominikTo:
Guessing from https://tools.ietf.org/html/rfc6350#section-6.2.5 I would say "no". Unfortunately we just see them in the wild every now and then.

evert:
Hm... kinda sucks.. makes me wonder if we should have rejected them from the start.

Issue #41, opened 2013-06-07
Property\Text should handle CHARSET encoding for VCARD v2.1


clue:
Now that we finally have a `Property::getValue()` (*hooray*) that returns the value in a datatype native to PHP without all the base64/quoted-printable encoding fuzz, I think it should also support (de)encoding according to its CHARSET parameter. Specifically MS outlook (and very likely others) usually uses the native system encoding and includes this in the CHARSET declaration only once special characters (outside of the standard ASCII range?) are included in the property value.

Now that VObject v3 is right around the corner, I'd vote for including it before the feature freeze to avoid breaking the API once again.

I'll see if I can contribute some patches, but I wanted to post this right away.

Hywan:
ping?

evert:
It's a feature request, and it's still open. Not sure what else I could say about this...

Hywan:
We have #43 opened. Is it ready for a review?

evert:
The patch seems a bit incomplete, but unfortunately I haven't been able to get more into it yet...

evert:
Sorry it took 2 years to fix this, but I finally committed a patch that will cause the parser to automatically decode properties that have `CHARSET` specified to UTF-8. Currently only ISO-8859-1 is supported, but if you need more encodings, let me know.

clue:
Thanks @evert, the patch looks good to me and is much appreciated!

> Currently only ISO-8859-1 is supported, but if you need more encodings, let me know.

The original patch in #43 explicitly tested against `CHARSET=Windows-1252`. This is what a German MS Outlook version exported, so I suppose it makes sense to also take Windows' code pages into account here.

evert:
I'll definitely add support for WIndows-1252 then!

PR #255, opened 2015-08-31
Added test for outlook encoding issue

DominikTo:
Rebased version of #215 to master.


evert:
In the future, please just do a forced push in the original PR. This way we create a lot of extra issues and the context gets lost.

evert:
Not really sure what to do with this ticket. At the moment we don't really change encodings for anything, we kind of keep everything as-is, including this. It would make sense to convert everything to UTF-8 during the parsing phase, but that's a larger task. Vobject4 would be the right time to do that though.

DominikTo:
Oh, is there a way to change the target branch in the meantime? I thought that's still not possible to change from e.g. "3.4" to "master".

DominikTo:
#41 and #43 might be related to this one.

evert:
I just took a look at this branch and updated it with master. I think it would be good to get a true original that demonstrates the issue. `QUOTED-PRINTABLE` no longer exists in vCard 3.0, so I find it a bit suspicious that it appears here. Is it possible that it was actually a vCard 2.1 object?

DominikTo:
The original card didn't contain a PRODID, so it's unfortunately unknown what created this.

evert:
Maybe it's not worth fixing then. I think we need some sort of reputable source / repeatable issue to really make it worth our while. I will close this ticket, and if we come across a popular application that produces something like this, we can always reconsider.

PR #258, opened 2015-10-14
Fix rdate in vcalendar

jabdoa2:
VCalendar::expand ignores RDATE. Fixed that. Additionally, fixed timezone issue in RDateIterator.


evert:
Looks like a good patch. Would need some unittests to ensure that this is behaving correctly.

jabdoa2:
I added tests for both bugs. Both fail on master but work on my branch.

evert:
Excellent patch. 10/10. Thank you!

PR #263, opened 2015-11-11
[WIP] Fixing minor issues

jakobsack:
This PR fixes some minor issues I found so far in the Sabre::VObject library.

### DIFF
--- a/lib/Parser/Json.php
+++ b/lib/Parser/Json.php
@@ -116,7 +116,7 @@ class Json extends Parser {
         return $this->root->createComponent(
             $jComp[0],
             array_merge($properties, $components),
-            $defaults = false
+            false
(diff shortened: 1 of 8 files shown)


lib/Parser/Json.php
```diff
@@ -116,7 +116,7 @@ function($jComp) use ($self) {
         return $this->root->createComponent(
             $jComp[0],
             array_merge($properties, $components),
-            $defaults = false
+            false
```
evert: I kind of like the extra `$defaults`. It doesn't do anything here, but it's a way to make a boolean flag self-documenting. Alternatively this could be replaced by a simple PHP comment.

evert:
Really great :+1:

PR #270, opened 2015-12-01
Support automatic decoding of Windows-1252

evert:


lib/Parser/MimeDir.php
```diff
@@ -460,6 +469,9 @@ protected function readProperty($line) {
                 case 'ISO-8859-1' :
```
staabm: Side note: what about iso-8859-15?

lib/Parser/MimeDir.php
```diff
@@ -460,6 +469,9 @@ protected function readProperty($line) {
...
                     $property['value'] = utf8_encode($property['value']);
                     break;
+                case 'Windows-1252' :
```
staabm: Some people call it cp1252, not sure you want to also handle this "alias"

evert: I basically just want to add the character sets on a case-by-case basis. ISO-8859-15 makes a lot of sense, but if there's no software out there that specifically marks files at ISO-8859-15, then I don't really see a strong need to add it.

Basically I just want to wait until people request these specific features.

staabm: Okidoki

PR #279, opened 2015-12-09
VCalendar: Fix timezones stripping

Hywan:
Fix https://github.com/fruux/sabre-vobject/issues/278.

The issue is that the `stripTimezones` closure is recursive but the
“self” is not defined. This way it works.

Minimal Working Example:

    $a = function ($i) use (&$a) {
        var_dump($i);

        if ($i < 10) {
            $a($i + 1);
        }
    };

Here we can re-use “self” (`$a`) inside the closure because it is closed
into the same lexical scope thanks to `use (&$a)`.

### DIFF
--- a/lib/Component/VCalendar.php
+++ b/lib/Component/VCalendar.php
@@ -299,7 +299,7 @@ class VCalendar extends VObject\Document {
             $timeZone = new DateTimeZone('UTC');
         }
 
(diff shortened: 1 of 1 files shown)


staabm:
I guess we should have a unit test for this path

DominikTo:
:+1: for the PR and :+1: for a unit test, that would have catched this. ;-)

PR #281, opened 2015-12-11
Rdate without time and rdate plus rrule

jabdoa2:


lib/Recur/EventIterator.php
```diff
@@ -66,6 +66,14 @@ class EventIterator implements \Iterator {
     protected $timeZone;
 
     /**
+     * We have to remember the previous time. Because every time should be only
+     * emitted once
+     *
+     * @var DateTimeZone
+     */
+    protected $previousEvent = false;
```
staabm: do we need the defined default `false` here

lib/Recur/EventIterator.php
```diff
@@ -387,19 +395,22 @@ function next() {
...
-                return;
-
             }
 
         }
 
-        $this->currentDate = $nextDate;
+        // 1. Always include dtstart (counter == 1)
+        // 2. Do not emit an event twice
+        if ($this->counter == 1 or ! $this->previousEvent or $nextDate != $this->previousEvent) {
```
staabm: AFAIR this project uses `||` for OR

jabdoa2:
Why did this build break?

evert:
Hi! Simply run `./bin/sabre-cs-fixer fix .` to automatically fix CS-related issues

tests/VObject/Recur/EventIterator/HandleRDateExpandTest.php
```diff
@@ -54,7 +54,125 @@ function testExpand() {
...
+UID:BDF720338B3EC8D2C09CE95CFF85C0B1-A26036CE0B219773
+DTSTAMP;VALUE=DATE-TIME:20151211T121402Z
+CREATED;VALUE=DATE-TIME:20151013T061627Z
+LAST-MODIFIED;VALUE=DATE-TIME:20151014T093742Z
+DTSTART;VALUE=DATE-TIME;TZID=Europe/Berlin:20151028T124500
+DTEND;VALUE=DATE-TIME;TZID=Europe/Berlin:20151028T141500
+SUMMARY:test summary
+DESCRIPTION:test desc
+SEQUENCE:1
+RDATE;VALUE=DATE:20151028,20151111,20151125,20151209,20160113,20160120,2016
```
evert: What's interesting with this example, is that I don't believe that we are _supposed_ to be able to support this particular syntax. As far as I know the value-type of `RDATE` needs to be identical to the value type of `DTSTART`. So if `DTSTART` is a `DATE-TIME`, then so must `RDATE` be...

Did you find this in the wild somewhere?

jabdoa2: Thats an ICS from our Kolab instance. According to the spec this is fine. Roundcube does properly support it. Thunderbird/Lightning also fails.

jabdoa2: See: http://www.kanzaki.com/docs/ical/rdate.html

evert: Well, if the syntax is supported by iCalendar, it would have a very specific meaning, and it's not "inherit the timestamp from the master event". The only way I could see that this should logically behave, is that subsequent events become `DATE`-only 'floating date' events.

While I agree that the spec does not appear to specifically forbid it, it also definitely does not suggest inheriting the time if only a date is specified in `RDATE`.

If Kolab is the only thing generating this, then I think that it's better to fix Kolab. I'm also curious if you can link me to the Lightning bugzilla ticket, as I somewhat suspect that Phillip will say the same thing.

jabdoa2:
We did not create a lightning bug, yet. Kolab is just using Roundcube and iRony so if this is a bug it should be in there somewhere but not specific to Kolab.

Is there any documentation about the semantics of RDATE?

evert:
rfc5545 is the only official reference. It's a bit sparse when it comes to talking about `RDATE`. So if Kolab did not generate this file, do you know what did?

evert:
It does say this though:

> The "DTSTART" property
      value SHOULD match the pattern of the recurrence rule, if
      specified.  The recurrence set generated with a "DTSTART" property
      value that doesn't match the pattern of the rule is undefined.

jabdoa2:
Kolab is a bunch of different programs which are well intergrated. The event was probably created by our secretary using Roundcube (provided by Kolab). We export this using iRony (provided by Kolab) to a Webinterface where we display it using SabreDav.

jabdoa2:
Looks like iRony is a product of Kolab and it uses SabreDav internally. So it might be another SabreDav bug?

evert:
So, excuse my ignorance, but are you saying that it's likely that Roundcube generated the event? If so, I think it would be better to report this as a Roundcube bug than a sabre/vobject bug.

Even if you can warp and bend the iCalendar standard to behave like you say it should behave, it would still be behavior that's unlikely to work correctly with a lot of systems. For the sake of interopability, it's better if Roundcube simply generates `RDATE` properties with their full `DATE-TIME` values.

evert:
Well, sabre/dav itself does not generate iCalendar objects, it just accepts whatever you throw at it. In the future we might enable stricter validation of input data, so at one point in the future we might simply reject these types of objects imply from ever reaching the server.

But there's still 'something else' that originally created that iCalendar object. That could be a web application, or a CalDAV client somewhere. It's also possible that iRony maps to an inferior internal data structure, and simply iRony's customizations might generate this data.

Does the iCalendar object you're using have a `PRODID` property? That might be a good hint as to who generated the object.

jabdoa2:
PRODID:-//Roundcube libcalendaring 1.1.3//Sabre//Sabre VObject 2.1.3//EN

evert:
Well then I would say Roundcube/libcalendaring is probably at fault.

jabdoa2:
https://git.kolab.org/T917

evert:
Thanks, I subscribed to the issue. So if they have any questions, I'll help. I'll close this PR now though.

PR #283, opened 2015-12-12
Using generators in sabre/vobject where they make sense

### LINKED ISSUE #282: Use generators where we return arrays now
This is probably the last vobject 4 change. Even though we're already in beta, I think it's worth implementing.

evert:
Fixes #282

### DIFF
--- a/lib/Cli.php
+++ b/lib/Cli.php
@@ -583,7 +583,7 @@ HELP
 
         };
 
-        $children = $vObj->children();
+        $children = iterator_to_array($vObj->children());
         $tmp = $children;
         uksort(
             $children,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ ChangeLog
   `Sabre\VObject\Settings::$maxRecurrences` and is also honored by the
   FreeBusyGenerator.
 * #287: `expand()` did not work correctly on events with sub-components.
+* #282: We're now returning generators from functions such as `select()`,
+  `getComponents()`, `getBaseComponents()` and `select()`.
 
 
 4.0.0-beta1 (2015-12-02)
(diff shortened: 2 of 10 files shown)


evert:
Benchmarks

Before

```
__global__  |||||||||||||||||||||||||||||||||||||||||||||||||||| 57830ms, 100.0%
parse                                                              257ms,   0.4%
fb          |||||||||||||||||||||||||||||||||||||||||||||||||||| 57573ms,  99.6%

peak memory usage: 14M
current memory usage: 13M
```

After

```
__global__  |||||||||||||||||||||||||||||||||||||||||||||||||||| 57561ms, 100.0%
parse                                                              256ms,   0.4%
fb          |||||||||||||||||||||||||||||||||||||||||||||||||||| 57305ms,  99.6%

peak memory usage: 15M
current memory usage: 13M
```

Unfortunately this implies there's nearly no performance benefit, and slightly more memory usage. Because of this, I'll discard the patch.

lib/Cli.php
```diff
@@ -583,7 +583,7 @@ protected function serializeComponent(Component $vObj) {
 
         };
 
-        $children = $vObj->children();
+        $children = iterator_to_array($vObj->children());
```
staabm: I guess this was the culprit of not getting it faster... Did you try ArrayIterator::uksort?

staabm: Same in the other places of iterator_to_array

evert: In my benchmark I mostly tested things that use things like property-access. A lot of statements like this `$vcalendar->VEVENT->DTSTART` use (in this PR) multiple generators.

I was hoping there would be a fair bit of win there. Changing the iterator_to_array and ksort statements only really help with serializing, which is something I haven't really tested for. I suspect though that the win will be minimal too...

PR #284, opened 2015-12-18
Include DTEND in cancellation emails

kewisch:
When receiving a CANCEL email, there is no dtend. We'd like to display the end date in the email template we are generating, so it would be nice to include this patch (on master obviously, but if possible we'd also like to have this backported to 3.4 in a way that sabre-dav makes use of it)


evert:
Hey @kewisch ,

Looks good and I'd be happy to backport this. To fix the build, run:

    ./bin/sabre-cs-fixer .

And add  `DTSTAMP:**ANY**` to the complaining tests. `**ANY**` is a 'magic' value used by the unittest.

kewisch:
Thanks for the hint. To be honest I created the patch against 3.4 and then ran `phpcs -p --standard tests/phpcs/ruleset.xml` and it didn't come up with errors. I'll need to set up a vm with a newer php version to do so. I might just fix the errors manually :)

evert:
Alright makes sense =)

kewisch:
There we go, sorry for the extra build notifications!

evert:
just did the backport to the 3.4 branch as well.

kewisch:
Thanks for the quick turnaround!

PR #287, opened 2016-01-07
PHP 7 support for vobject 3.4

DeepDiver1975:
@evert this will bring php7 to vobject 3.4 - no idea if the class rename is valid in terms of api breackage - worth a 3.5???

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ php:
 matrix:
   fast_finish: true
   allow_failures:
-    - php: 7
+    - php: hhvm
 
 script:
   - phpunit --configuration tests/phpunit.xml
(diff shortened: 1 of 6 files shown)


staabm:
I guess vobject 4.x was created with the BC breaking changes for that reason.

Doing this BC in a 3.4 -> 3.5 would be non semver compilant.

evert:
Yea the issue was indeed that I didn't want to make that break in a 3.x. But I think in this case the benefits may outweigh the drawbacks. Luckily I think that very little people will rely on the class names that are being renamed here.

The only people that fall in that category, are the ones that define their own INTEGER/FLOAT properties or that needed alternative behavior for the existing ones. I think that group of people probably approximates 0.

So lets break the rules for this one and release this as 3.5.

DeepDiver1975:
Thanks a lot, @evert

PR #291, opened 2016-01-13
Trim whitespaces from the value of the property, except the description

laoneo:
Ical content from some event providers do contain white spaces after some properties like the DTSTAMP field. This is not valid ical content, but can be easily solved by trimming white spaces from the value of the property. It makes VObject more robust.

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -93,6 +93,18 @@ abstract class Property extends Node {
      */
     function setValue($value) {
 
+        if ($this->name != 'DESCRIPTION') {
+            if (is_string($value)) {
+                $value = trim($value, ' ');
+            }
+            if (is_array($value)) {
+                foreach ($value as $key => $val) {
+                    if (is_string($val)) {
+                        $value[$key] = trim($val, ' ');
+                    }
+                }
+            }
+        }
         $this->value = $value;
 
     }
--- a/tests/VObject/PropertyTest.php
+++ b/tests/VObject/PropertyTest.php
@@ -48,6 +48,31 @@ class PropertyTest extends \PHPUnit_Framework_TestCase {
 
     }
 
+    function testSetValueWithWhiteSpaces() {
+
+        $cal = new VCalendar();
+
+        $property = $cal->createProperty('propname', 'propvalue');
+        $property->setValue(' value2 ');
+
+        $this->assertEquals('PROPNAME', $property->name);
+        $this->assertEquals('value2', $property->__toString());
+
+    }
+
+    function testSetArrayValueWithWhiteSpaces() {
+
+        $cal = new VCalendar();
+
+        $property = $cal->createProperty('propname', 'propvalue');
+        $property->setValue(array('value2 '));
+
+        $this->assertEquals('PROPNAME', $property->name);
+        $this->assertEquals('value2', $property->__toString());
+
+    }
+
+
     function testParameterExists() {
 
         $cal = new VCalendar();


staabm:
Your tests only test for trailling slashes but your fix also trimms leading ones

laoneo:
Adapted the test to test also for leading spaces.

evert:
What problem does this aim to solve? I find it a bit hard to believe that people accidentally adding a space when setting a property such as this:

$event->DTSTAMP = ' 20160112T000000';

is a common enough occurrence that it warrants fixing. If anything I would rather see `setValue()` being strictly validated for correctness, which would be a bit more helpful I think (although it does slow things down a bit).

Can you perhaps describe the situation that led you to wanting to add this feature?

evert:
Never mind. It's late, and I looked past your initial comment here:

> cal content from some event providers do contain white spaces 

In that case the first thing I would suggest is to try to get that provider to fix the iCalendar they produce. The bug is with theirs. Who is it?

Aside from that then, the best course of action is still to try to actually make vobject more strict, not less. Instead of stripping the whitespace, we really should outright reject it. Thoughts?

laoneo:
I totally agree that event providers need to provide correct ical content and it is not up to VObject to correct it. But reality is always different that theory. The client of mine does fetch events from 
http://ical.avaibook.com/ical/ua_051ac4c08a187ec0167514269bd42c2c-1c9ac0159c94d8d0cbedc973445af2da.ics
The DTSTAMP and UID fields do have spaces after the value.

evert:
> But reality is always different that theory. 

Is it though? =) Have you reported the bug at availbook already, or did you go here first? After all, since it's their bug it would make sense to get them to fix it...

laoneo:
So, no chance this patch gets included?

evert:
I'll ask again:

have you reported this bug at availbook first? Since it's their bug, it would make sense to get them to fix it.

This is not unlike a website that has a mistake in their HTML, that would cause browsers to render it incorrectly. Would you report the bug at Mozilla or Google Chrome, or at the owners of the website?

laoneo:
No, because the problem happens on a client of mine, so basically I'm only the middle man. I will give them a hint, but honestly I'm not really convinced they will fix it. What happens if they say no?

evert:
If they say no then come back with the answer so we can think of a possible workaround.

I put a lot of time and effort in this project, the support I give here is mostly free.

Please consider that asking me to build a workaround for a bug for a different project without as much asking the other project first to fix it is quite insulting.

It tells me that you assume that my time is worth less than theirs. Step 1 should be to report the bug where it belongs, if they refuse to fix it or answer you, then I will donate my free time to help you out with this specific non-standard edge-case.

laoneo:
I guess you did understand me wrong, I don't think that your time is worth less than theirs or somebody else time. It is just an experience from me, that companies do not fix stuff fast which is not core business of them. I really appreciate your work here, and wanted to pay already for a business case I needed from fruux, but they told me they are not interested at the moment :-)
Anyway, we are talking here about a PR, so all the work is already done, you need to push only the merge button, because I didn't want to waste your valuable time.
I opened a ticket on their web site, we will see what they say. Again, I do really appreciate the work you do here with sabre!!

staabm:
@laoneo do they have a www accessible bug tracker you could crosslink here?

regarding the costs of this PR and regarding code lines. the actual cost for mainting this kind of hack start _after_ clicking the merge button.. therefore to choose wisely is very important.

laoneo:
Making a parser more robust is not a hack more of an improvement.
Ticket can be found here http://www.avaibook.com/soporte/ticket.php?track=W47-QJN-VHYT.

evert:
> Making a parser more robust is not a hack more of an improvement.

That's where we kind of disagree. There's a lot of vobject users, and allow bad iCalendar data through means for example that caldav servers now accept bad iCalendar data, and that data will be sent back to other clients that may not support it. It actually reduces the robustness, and therefore I absolutely need to do the due diligence to make sure it's an absolutely needed change.

If it does indeed turn out that nothing can be done at availbook's end, I'll need to approach this problem a bit differently from this patch. I want to make sure that it's not a behavior that's enabled by default, and instead I would like to make it part of the repair & validation system. This means that after we add this change, you'd have to call something like:

    $vcalendar->repair(Node::OPTION_REPAIR);

To automatically adjust the broken values. This option already does quite a few things to make broken data better, so it's a more appropriate place. This way we can also target the specific node-types (`DATE-TIME` and `DATE`) instead of everything except `DESCRIPTION`.

Thanks for posting that ticket. Fingers crossed there's a positive response.

laoneo:
Wasn't aware about the repair and validation system. Then I agree, the fix needs to be done there. Can you give me a link to the repair code?

evert:
Every property and component class has it's own `validate()` function. Since you specifically talked about `DTSTAMP`, this would be this method in this class:

https://github.com/fruux/sabre-vobject/blob/master/lib/Property/ICalendar/DateTime.php#L378

Here's an example of another `validate()` function that looks for the `REPAIR` option and fixes the problem if possible:

https://github.com/fruux/sabre-vobject/blob/master/lib/Component/VCard.php#L224

The easiest way to do a quick test for this, is by using the cli tool. If you checked out this repository and ran `composer install`, you should be able to run it as such:

    bin/vobject validate [filename.ics]

If you're able to write this patch (specific to `DATE-TIME` values) and supply a unittest that validates your change, I'll _definitely_ merge your patch, regardless of if avaibook responds/cares =)

PR #295, opened 2016-02-04
fix infinite iteration when rrule YEARLY and BYMONTH empty AND fix fatal error when DSTART is not set

cedtaz:
Lets not have infinite iteration when you have this kind of rule : 
"RRULE:FREQ=YEARLY;INTERVAL=1;BYMONTH="

Indeed, when we have a rrule yearly with a parameter bymonth empty, the function nextYearly() iterate infinitely like we can see here (line 605) :

do {
$currentMonth++;
if ($currentMonth > 12) {
$currentYear += $this->interval;
$currentMonth = 1;
}
} while (!in_array($currentMonth, $this->byMonth));

### DIFF
--- a/lib/Component/VEvent.php
+++ b/lib/Component/VEvent.php
@@ -58,6 +58,9 @@ class VEvent extends VObject\Component {
 
         }
 
+        if (!isset($this->DTSTART)) {
+            return false;
+        }
         $effectiveStart = $this->DTSTART->getDateTime($start->getTimezone());
         if (isset($this->DTEND)) {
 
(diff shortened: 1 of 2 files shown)


evert:
Interesting. I' thinking we should probably treat an empty BYMONTH as not being set at all instead though. I'm also thinking that we maybe should just throw an exception as it might be invalid.

cedtaz:
I had also thought of putting an exception but the problem is the validation of the object (with the validate() method), it returns us no warnings...

cedtaz:
I had a second commit in order to fix fatal error when DSTART is not set

PR #300, opened 2016-02-17
[RFC] add VCard::getByType()

kbond:
Hello! I had a requirement for this feature so I thought I would create a PR to discuss.

### DIFF
--- a/lib/Component/VCard.php
+++ b/lib/Component/VCard.php
@@ -411,6 +411,25 @@ class VCard extends VObject\Document {
     }
 
     /**
+     * Returns a property with a specific TYPE value (ADR, TEL, or EMAIL).
+     *
+     * This function will return null if the property does not exist. If there are
+     * multiple properties with the same TYPE value, only one will be returned.
+     *
+     * @param string $propertyName
+     * @param string $type
+     *
+     * @return VObject\Property|null
+     */
+    function getByType($propertyName, $type) {
+        foreach ($this->select($propertyName) as $field) {
+            if (isset($field['TYPE']) && $field['TYPE']->has($type)) {
+                return $field;
+            }
+        }
+    }
+
+    /**
      * This method should return a list of default property values.
      *
      * @return array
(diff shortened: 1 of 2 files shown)


lib/Component/VCard.php
```diff
@@ -411,6 +411,24 @@ function preferred($propertyName) {
     }
 
     /**
+     * Returns a specific type of multi fields (ADR, TEL, or EMAIL).
```
evert: Perhaps:

```
Returns a property with a specific TYPE value (ADR, TEL, or EMAIL)

This function will return null if the property does not exist. If there are multiple properties with
the same TYPE value, only one will be returned.
```

lib/Component/VCard.php
```diff
@@ -411,6 +411,24 @@ function preferred($propertyName) {
...
+     *
+     * @param string $propertyName
+     * @param string $type
+     *
+     * @return VObject\Property|null
+     */
+    function getForType($propertyName, $type) {
```
evert: I would suggest renaming this to `getByType` to be more in line with other functions.

lib/Component/VCard.php
```diff
@@ -411,6 +411,24 @@ function preferred($propertyName) {
...
+        foreach ($this->select($propertyName) as $field) {
+            if (isset($field['TYPE']) && $field['TYPE']->getValue() === $type) {
```
evert: What happens in a case when there's a property with multiple TYPE parameters, or a multi-value TYPE parameter? I think Parameter objects have a `->has` function that should be used instead.

lib/Component/VCard.php
```diff
@@ -411,6 +411,24 @@ function preferred($propertyName) {
...
+                return $field;
+            }
+        }
+
+        return null;
```
evert: This line is not needed, as by default it will return `null`.

lib/Component/VCard.php
kbond: Sure, what do you suggest?

lib/Component/VCard.php
kbond: something like:

```php
if (isset($field['TYPE']) && $field['TYPE']->has($type)) {
    // ...
}
```

kbond:
Ok, I have made the suggested changes.

PR #304, opened 2016-03-03
Issue 303 - Use number instead of string for COUNT property in RRULE

### LINKED ISSUE #303: jCal use string for count rrule property instead of number
The [jCal rfc](https://tools.ietf.org/html/rfc7265) specify that COUNT property should be a number in not a string.
But Sabre send this property as a String and fail when parsing jCal with COUNT as a number.

strokyl:
Fixes #303

### DIFF
--- a/lib/Property/ICalendar/Recur.php
+++ b/lib/Property/ICalendar/Recur.php
@@ -54,7 +54,7 @@ class Recur extends Property {
                     if (strcmp($k, 'until') === 0) {
                         $v = strtr($v, [':' => '', '-' => '']);
                     }
-                } else {
+                } elseif (is_array($v)) {
                     $v = array_map('strtoupper', $v);
                 }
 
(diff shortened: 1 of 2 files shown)


lib/Property/ICalendar/Recur.php
```diff
@@ -54,7 +54,7 @@ function setValue($value) {
                     if (strcmp($k, 'until') === 0) {
                         $v = strtr($v, [':' => '', '-' => '']);
                     }
-                } else {
+                } else if (!is_int($v)) {
```
staabm: I guess the logic expects a `is_string`?

strokyl: Hum if I am not wrong no it's is_int($v), is_string case is handle before, this else if(!is_int($v)) will fail if $v is a int (and not an array) for the int case, there nothing to do just store the value which is done outside the if else if.

staabm: you are right, `is_string` is wrong...  `is_array` it is ;-)

lib/Property/ICalendar/Recur.php
```diff
@@ -54,7 +54,7 @@ function setValue($value) {
                     if (strcmp($k, 'until') === 0) {
                         $v = strtr($v, [':' => '', '-' => '']);
                     }
-                } else {
+                } elseif (!is_int($v)) {
```
evert: To add to that previous conversation, it might simply be a bit better to use `is_array` and not `is_int`. I think it makes this a bit clearer. Other than that, patch looks good!

strokyl: Yes it's better like that, I have fixed it. 
Thank you.

PR #305, opened 2016-03-09
Invitee change status in first event of a recurring events.

vkomrakov-sugar:
Fixed a invitee change status in the first event and a value definition "recurrence-id" with a timezone.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -345,7 +345,7 @@ class Broker {
 
         // Finding all the instances the attendee replied to.
         foreach($itipMessage->message->VEVENT as $vevent) {
-            $recurId = isset($vevent->{'RECURRENCE-ID'})?$vevent->{'RECURRENCE-ID'}->getValue():'master';
+            $recurId = isset($vevent->{'RECURRENCE-ID'}) ? $vevent->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() : 'master';
             $attendee = $vevent->ATTENDEE;
             $instances[$recurId] = $attendee['PARTSTAT']->getValue();
             if (isset($vevent->{'REQUEST-STATUS'})) {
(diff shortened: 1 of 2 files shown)


evert:
Hi @vkomrakov-sugar ,

The 3.4 branch is effectively dead. Can you target `master` or `3.5` instead?

evert:
Unless you really want another 3.4.x release though, I definitely could do that, but I think 3.5 or 4.0 will probably work for you. let me know ;)

Michael-Gusev:
Hi @evert. Problem for us that we can't use php 5.5, only 5.4 and because of that we stay on sabre 3.0.x.
Is it okay to do PR against 3.5 branch and get those changes in sabre 3.0.x?

evert:
In that case, the 3.5 is best. This PR was originally targetting the 3.4 branch.

lib/ITip/Broker.php
```diff
@@ -345,7 +345,7 @@ protected function processMessageReply(Message $itipMessage, VCalendar $existing
 
         // Finding all the instances the attendee replied to.
         foreach($itipMessage->message->VEVENT as $vevent) {
-            $recurId = isset($vevent->{'RECURRENCE-ID'})?$vevent->{'RECURRENCE-ID'}->getValue():'master';
+            $recurId = isset($vevent->{'RECURRENCE-ID'}) ? $vevent->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() : 'master';
```
evert: I'm curious why getTimestamp() was used instead of getValue(). The 'value' should always refer to the value in the timezone of the original event, so I don't believe there's a risk for timezone-related issues here.

lib/ITip/Broker.php
```diff
@@ -404,8 +404,12 @@ protected function processMessageReply(Message $itipMessage, VCalendar $existing
 
                 $newObject = $recurrenceIterator->getEventObject();
                 $recurrenceIterator->next();
-
-                if (isset($newObject->{'RECURRENCE-ID'}) && $newObject->{'RECURRENCE-ID'}->getValue()===$recurId) {
+                if (isset($newObject->{'RECURRENCE-ID'}) && $newObject->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() === $recurId) {
+                    $found = true;
+                } elseif ($recurrenceIterator->key() === 1 && $newObject->DTSTART->getDateTime()->getTimestamp() === $recurId) {
```
evert: Could you explain this line? `key()` in particular strikes as a bit odd. Wouldn't you want to refer to the first event? In addition to that, there's no guarantee that the first event is the master event. It's possible to make a recurrence and create an exception for the second instance, and set the DTSTART for that _before_ the master event.

Maybe I simply don't fully understand this change =)

evert:
I commented in the wrong PR. Ignore these

PR #312, opened 2016-04-01
incorrect work with timezone without event master

vkomrakov-sugar:


Michael-Gusev:
@evert could you review?
Problem here that `$recurId` has different values / format in different places:
in `processMessageReply` method it's `YmdTHisZ` but in `$existingObject->VEVENT` так `YmdTHis`.

evert:
The problem with this proposed fix is that it assumes that the first `VTIMEZONE` holds the correct info, which might very well be incorrect. I'm looking a bit further into this to see if there's another fix possible.

Thanks for the complete unittest though. Very awesome.

Michael-Gusev:
Thank you, @evert. Could you tell us next release date?

evert:
Yesterday!

Michael-Gusev:
Ahahaha :+1:

Michael-Gusev:
Hi @evert, one more thing. We can't use 5.5 and use 5.4 but I found fix was released under 4.1.0, and we need it under 3.5.x.
Could you add it to 3.5.x release?
Thank you.

Michael-Gusev:
@evert, could you check my prev comment?

evert:
I just backported it to the 3.5 branch.

PR #326, opened 2016-06-29
Fix #316: strip alarms when necessary in itip broker

### LINKED ISSUE #316: Feature request: strip VALARM from invitations
Following https://github.com/fruux/sabre-dav/issues/819.

The goal is here is to have VAlarms which stay private to each attendee and to the organizer.
So Valarms should not be propagated to attendee ICS when the organizer adds or modify a VAlarm for himself/herself.

I maybe can do this fix myself if I have time but I am not sure of where it should take place. I am guessing `sabre-vobject/lib/ITip/Broker.php` but I do not know the project that well for now...

GrahamLinagora:
It does not totally fix the problem but I would like your opinion on this because I do not know how to go further.

For now, if an event has a creator and an attendee:
- creator adds an alarm: the attendee copy of thenevent has no alarm
- attendee adds an alarm: the creator event still has its previous alarm and the attendee alarm is correct
- creator modifying his alarm erases the attendee alarm

The problem is that I strip VAlarms from Itip messages. I think the good thing to do to prevent this case is to not send any message if only the Valamr has changed. Is there any Vevent comparison tool for that in vobject ?

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -2,6 +2,7 @@
 
 namespace Sabre\VObject\ITip;
 
+use Sabre\VObject\Component;
 use Sabre\VObject\Component\VCalendar;
 use Sabre\VObject\DateTimeParser;
 use Sabre\VObject\Reader;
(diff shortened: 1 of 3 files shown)


lib/ITip/Broker.php
```diff
@@ -572,6 +583,9 @@ protected function parseEventForOrganizer(VCalendar $calendar, array $eventInfo,
                 foreach ($attendee['newInstances'] as $instanceId => $instanceInfo) {
 
                     $currentEvent = clone $eventInfo['instances'][$instanceId];
+
+                    $this->stripVAlarms($currentEvent);
```
evert: This can be shortened to:

    unset($currentEvent->VALARM);

And then you don't need that extra function

PR #327, opened 2016-07-08
Throw InvalidDataException if DateTimeImmutable constructor fails

rsto:
DateTimeParser fails for syntactically correct, but invalid datetimes with a generic PHP exception instead of InvalidDataException.

This e.g. causes sabre-dav's CalDAV plugin to fail with 500 Internal Server error for such datetimes, instead of rejecting the erroneous iCalendar payload.

For example, a datetime in the 25th hour of the day passes the regex validation but fails with \Exception

    DateTimeImmutable::__construct(): Failed to parse time string (2010-03-16 25:14:05) at position 11

Similarly, a date in the 13th month fails with

    DateTimeImmutable::__construct(): Failed to parse time string (2010-13-31) at position 6

This change fixes both and adds unit tests.

### DIFF
--- a/lib/DateTimeParser.php
+++ b/lib/DateTimeParser.php
@@ -43,7 +43,12 @@ class DateTimeParser {
         if ($matches[7] === 'Z' || is_null($tz)) {
             $tz = new DateTimeZone('UTC');
         }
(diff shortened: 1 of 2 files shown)


evert:
Can you re-base from master? Should fix the build. Thanks for the complete patch!

rsto:
Done. Travis checks passed now.

evert:
Thank you!

PR #330, opened 2016-07-14
Fix #330: Fixed case when overridden recurring events have same start date.

afedyk-sugarcrm:
If ```Sabre\VObject\Recur\EventIterator``` takes on input list os recurring events with same date start. In ```EventIterator::rewind``` generate overridden events index by timestamp of date start(```EventIterator::$overriddenEventsIndex```). So index offset will be rewritten by last event with same date start. 
Fix issue: #331

### DIFF
--- a/lib/Recur/EventIterator.php
+++ b/lib/Recur/EventIterator.php
@@ -328,7 +328,7 @@ class EventIterator implements \Iterator {
         $index = array();
         foreach($this->overriddenEvents as $key=>$event) {
             $stamp = $event->DTSTART->getDateTime($this->timeZone)->getTimeStamp();
-            $index[$stamp] = $key;
+            $index[$stamp][] = $key;
         }
         krsort($index);
         $this->counter = 0;
@@ -375,8 +375,9 @@ class EventIterator implements \Iterator {
         // overridden event may cut ahead.
         if ($this->overriddenEventsIndex) {
 
-            $offset = end($this->overriddenEventsIndex);
+            $offsets = end($this->overriddenEventsIndex);
             $timestamp = key($this->overriddenEventsIndex);
+            $offset = end($offsets);
             if (!$nextDate || $timestamp < $nextDate->getTimeStamp()) {
                 // Overridden event comes first.
                 $this->currentOverriddenEvent = $this->overriddenEvents[$offset];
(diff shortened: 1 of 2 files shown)


lib/Recur/EventIterator.php
```diff
@@ -328,7 +328,10 @@ public function rewind() {
         $index = array();
         foreach($this->overriddenEvents as $key=>$event) {
             $stamp = $event->DTSTART->getDateTime($this->timeZone)->getTimeStamp();
-            $index[$stamp] = $key;
+            if (!array_key_exists($stamp, $index)){
```
Michael-Gusev: useless if `$index[$stamp][]` will work even if `$index[$stamp]` doesn't exist.

lib/Recur/EventIterator.php
```diff
@@ -386,7 +390,10 @@ public function next() {
                 $this->currentDate = $this->currentOverriddenEvent->DTSTART->getDateTime($this->timeZone);
 
                 // Ensuring that this item will only be used once.
-                array_pop($this->overriddenEventsIndex);
+                array_pop($this->overriddenEventsIndex[$timestamp]);
+                if (empty($this->overriddenEventsIndex[$timestamp])){
```
Michael-Gusev: follow code standards.

Michael-Gusev: no reason to use empty, `!` is enough.

tests/VObject/Recur/EventIterator/SameDateForRecurringEventsTest.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Class SameDateForRecurringEventsTest
+ */
+class SameDateForRecurringEventsTest extends \PHPUnit_Framework_TestCase
+{
+
+    /**
+     * Checking is all events iterated by EventIterator.
+     */
+    public function testIsAllEventImported()
```
Michael-Gusev: testAllEventsArePresentInIterator

evert:
Thank you, looks like a good patch! I appreciate the attention to detail

evert:
I just realized that this change was targetting the 3.3 branch, which is dead. I'm reverting the changes and re-applying it to 3.5.

Michael-Gusev:
Hi @evert, we use sabre 3.0 because php 5.4 and it points to vobject 3.3, is it possible to have fix and replase for 3.3 too?

evert:
sabre/dav 3.0.x does in fact work with sabre/vobject 3.5.x

Michael-Gusev:
Hm... but composer file say about 3.3.x
https://github.com/fruux/sabre-dav/blob/3.0/composer.json#L18

evert:
Thats not what that means. It includes every version lower than 4.0.0

Michael-Gusev:
Got it, thank you!

evert:
I was on the phone so I was a bit brief... but basically the `^` denotes that it uses semantic versioning, and with semantic versioning you can assume that given version x.y.z, that every version after that and before x+1.0.0 will work. You are probably confusing the `^` with `~`, which works differently.

Michael-Gusev:
@evert  thank you for info.

Michael-Gusev:
when might we expect release for 3.5.3?

Issue #329, opened 2016-07-12
Endless loop in RRuleIterator if BYMONTHDAY does not exist in BYMONTH


PHPGangsta:
I'm using vobject 4.1.0

I have the following file:
```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//something//DE
CALSCALE:GREGORIAN
X-WR-TIMEZONE:Europe/Berlin
BEGIN:VEVENT
UID:20160103T123422CET-9863LIMv8E
DTSTAMP:20160103T113422Z
DESCRIPTION:important date
DTSTART;TZID=Europe/Berlin:20151231T000000
DTEND;TZID=Europe/Berlin:20151231T235900
RRULE:FREQ=YEARLY;COUNT=6;BYMONTHDAY=31;BYMONTH=11
SUMMARY:important date
TRANSP:OPAQUE
END:VEVENT
END:VCALENDAR
```

The problem here is the BYMONTHDAY=31;BYMONTH=11. It's invalid, there is no day 31 in November. But there is no Exception thrown, and my script (in this case the migration17.php script) runs into an endless loop. You can reproduce it if you do the following:
```php
$vObject = \Sabre\VObject\Reader::read($calendarData);
$it = new \Sabre\VObject\Recur\EventIterator($vObject, '20160103T123422CET-9863LIMv8E');
```
The problem seems to be in line 558:
https://github.com/fruux/sabre-vobject/blob/master/lib/Recur/RRuleIterator.php#L558
$occurrences is always empty, that's why the "while(true)" runs forever. Why is it empty? Because "out of range dates" are skipped in line 825
https://github.com/fruux/sabre-vobject/blob/master/lib/Recur/RRuleIterator.php#L825

I guess that's causing the endless loop.

Not sure how to fix the problem. Maybe the "out of range date" detection should be earlier, before entering the "while(true)" loop? Or at least stop the loop when $currentYear reaches year 3000 or so?

willrowe:
I found another case where the `RRULE` is not being validated properly. A validation step should probably be added before iterating to ensure that it is actually iterable.

If for whatever reason a frequency isn't provided, it will loop until it hits the limit. As an example, the following `RRULE` can be passed and iterated without any exceptions until it hits the limit:

```
BEGIN:VEVENT
RRULE:INTERVAL=1;BYMONTH=8;BYMONTHDAY=8
DTSTART:20220807T210000Z
END:VEVENT
```

PR #332, opened 2016-07-14
fixes issue #329 (Endless loop in RRuleIterator if BYMONTHDAY does no…

PHPGangsta:
…t exist in BYMONTH)

### DIFF
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -588,6 +588,11 @@ class RRuleIterator implements Iterator {
                     (int)$currentDayOfMonth
                 );
 
+                // maximum year 10000 should be safe to stop (2016 + max 3500 yearly occurrences)
+                if ((int)$currentYear > 10000) {
+                    throw new NoInstancesException('This recurrence rule does not generate any valid instances, stopped in year 10000');
+                }
+
             }
 
             // If we made it here, it means we got a valid occurrence
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -8,6 +8,7 @@ use Iterator;
 use Sabre\VObject\DateTimeParser;
 use Sabre\VObject\InvalidDataException;
 use Sabre\VObject\Property;
+use Sabre\VObject\Settings;
 
 /**
  * RRuleParser.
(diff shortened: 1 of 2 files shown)


lib/Recur/RRuleIterator.php
```diff
@@ -588,6 +588,11 @@ protected function nextYearly() {
                     (int)$currentDayOfMonth
                 );
 
+                // maximum year 10000 should be safe to stop (2016 + max 3500 yearly occurrences)
+                if ((int)$currentYear > 10000) {
```
evert: We have a setting for this, that might be a bit better:

https://github.com/fruux/sabre-vobject/blob/master/lib/Settings.php#L37

lib/Recur/RRuleIterator.php
```diff
@@ -588,6 +588,11 @@ protected function nextYearly() {
...
+                    throw new NoInstancesException('This recurrence rule does not generate any valid instances, stopped in year 10000');
```
evert: I'm also not sure if an exception is really correct. The standard says that if a recurrence rule generates dates that are invalid (february 29th for some years for example) then those instances should just be skipped. So it's not necessarily "invalid" even if it's actually pretty wrong.

I think if we use the MAX_DATE setting we should simply just exit the loop and don't generate additional instances. That brings it in line with how the rest of the standard behaves I believe.

PHPGangsta: You might be right... But how do we leave the loop here without returning a "$this->currentDate"? Do we have to set it to NULL and then return?

evert: Good point! Maybe we do need an exception but then catch it elsewhere. The goal ultimately is that the `valid()` function needs to return false for these cases.

PHPGangsta: Any new ideas here? The endless loop is still present, it's a potential DoS attack vector. If running in a webserver, the request takes 100% CPU and high memory. It's killed hopefully after max 30 seconds, but 10 or 20 parallel requests that trigger this endless loop is a serious problem.
If using the VObject parser in a Cronjob or some other CLI script (like the Migration script), without a script timeout limit, it will never end.

We should fix this problem. Please tell me where to catch the Exception and how to change the valid() function, or maybe do it yourself directly, might be easier ;-)

evert:
Hi! I just took another loop at this problem, and I took a slightly different approach to fixing it. I should have used your tests though, they're definitely cleaner.

PR #338, opened 2016-07-15
Some checks added for RRULE BYMONTH in Recur->validate()

PHPGangsta:


PHPGangsta:
What do you think, should BYMONTH <1 be repaired to 1, and BYMONTH > 12 repaired to 12? 

Or should the value (if there are multiple months) or the whole BYMONTH be removed if <1 or >12?

evert:
I think in cases like that there's no reasonable assumption we can make about what the client actually wanted to do. So I would be just as happy with no repair rule at all and just always let the server throw an error. Ideally these clients need to be fixed.

PR #345, opened 2016-08-23
Add support for stripping offset from beginning of identifiers

jpirkey:
fixes fruux/sabre-vobject#344


jpirkey:
Weird I didn't change any of the files TravisCI is complaining about.

jpirkey:
@evert is this something that I have to deal with?  It looks like TravisCI is complaining about php-cs-fixer complaining about files I didn't change.  Not sure where to go from here.

evert:
Hi @jpirkey ,

Sorry about that. The cs dependencies change over time, and get a bit more accurate. So the violation is totally on our end. I just fixed this on master, so if you merge from master (or rebase) the problems should get resolved automatically.

jpirkey:
No worries @evert !  Just wanted to make sure it wasn't something I needed to do.

Cheers

PR #346, opened 2016-08-25
fixed copyright info in timezone generator

staabm:
justed noticed because https://github.com/fruux/sabre-vobject/commit/b03193a7d32d18b2b56a1a1c222906ea58c8e016 contains a "oudated" copyright year in the generated file.


DominikTo:
Thanks!

PR #350, opened 2016-10-24
Catching another infinite loop condition.

### LINKED ISSUE #329: Endless loop in RRuleIterator if BYMONTHDAY does not exist in BYMONTH

evert:
Fixes #329. Thanks for the research @PHPGangsta

### DIFF
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -620,11 +620,13 @@ class RRuleIterator implements Iterator {
 
         }
 
-        $currentMonth = $this->currentDate->format('n');
-        $currentYear = $this->currentDate->format('Y');
-        $currentDayOfMonth = $this->currentDate->format('j');
+        //$currentMonth = $this->currentDate->format('n');
+        //$currentYear = $this->currentDate->format('Y');
+        //$currentDayOfMonth = $this->currentDate->format('j');
 
         $advancedToNewMonth = false;
+
+        $noOccurrenceCounter = 0;
 
         // If we got a byDay or getMonthDay filter, we must first expand
         // further.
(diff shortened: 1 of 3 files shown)


lib/Recur/RRuleIterator.php
```diff
@@ -620,19 +620,33 @@ protected function nextYearly() {
 
         }
 
-        $currentMonth = $this->currentDate->format('n');
-        $currentYear = $this->currentDate->format('Y');
-        $currentDayOfMonth = $this->currentDate->format('j');
+        //$currentMonth = $this->currentDate->format('n');
+        //$currentYear = $this->currentDate->format('Y');
+        //$currentDayOfMonth = $this->currentDate->format('j');
```
DominikTo: IF L623-L625 are no longer needed, they should be removed before merging.

alecpl:
This should be closed in favor of #416.

lib/Recur/RRuleIterator.php
Yescase: Gotcha

PR #364, opened 2017-01-24
Component's validation can now deduplicate properties. Fixes #363

### LINKED ISSUE #363: Component's validation should be able to remove duplicate properties if REPAIR is asked
Hey guys,

At [Linagora](https://www.linagora.com) we're currently working on interoperability issues for our [OpenPaas](https://github.com/linagora/openpaas-esn) product (which integrates Sabre as the CalDAV and CardDAV backend).

The last in date is that Yahoo calendar sends invites with two STATUS:CONFIRMED properties. While this is indeed invalid as per [RFC5445](https://tools.ietf.org/html/rfc5545#section-3.8.1.11), this is accepted by all other providers we tested.

I'm opening a PR to support deduplicating identical properties (same key, same value, same parameters) when REPAIR is asked as an option to [Component->validate()](https://github.com/fruux/sabre-vobject/blob/master/lib/Component.php#L607).

Regards,
David

ddolcimascolo:


evert:
This is a neat addition. Well done!

PR #370, opened 2017-03-09
Update VAlarm.php

dehart:
When the related property of a VALARM in a VTODO is "START" it should be Related to the DTSTART and this property MUST be present according to spec:
https://tools.ietf.org/html/rfc5545#section-3.8.6.3


evert:
The change is good, but the tests also need to pass

tpayen:
The change is good, but the test is bad, the ICS isn't RFC 5545 compliant

phil-davis:
Fixed (with tests) in PR #616

PR #372, opened 2017-03-24
For a vtodo, a start alarm must refer to dtstart

tpayen:
Hi,
Like the RFC 5545 (https://tools.ietf.org/html/rfc5545#page-74) said:
"In an alarm set to trigger on the "START" of an event or to-do,
      the "DTSTART" property MUST be present in the associated event or
      to-do."
So when related = start, we must use the dtstart of the vtodo, not the due. Since it's the same thing as the vevent, the if isn't needed.


tpayen:
woops i didn't see pr #370 sorry

PR #373, opened 2017-04-03
Pin PHPUnit version to avoid breaking unit tests for PHP 7+

PowerKiKi:
Because PHPUnit recently released a v6 with very important breaking changes, it is required to specify a version to not break our tests. Typically this will solve broken tests as seen in #371.

Also see [this excellent explanation](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) as to why composer.lock should be committed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,8 +37,8 @@
         "sabre/xml"    : ">=1.5 <3.0"
     },
     "require-dev" : {
-        "phpunit/phpunit" : "*",
-        "sabre/cs"        : "^1.0.0"
+        "sabre/cs"        : "^1.0.0",
+        "phpunit/phpunit": "^4.8"
 
     },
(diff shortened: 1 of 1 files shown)


PowerKiKi:
I noticed that PHP 5.5 is still supported by this project even though it [reached end of life 8 months ago](http://php.net/eol.php). I'd strongly suggest getting rid of that support and update to PHPUnit 5.7.

Or even consider ditching PHP 5.6 altogether and stick to only 7+ as [PHPUnit did themselves 8 months ago](https://github.com/sebastianbergmann/phpunit/commit/6459e984793d8a9e7a987ee07318b3759aa561ad).

tbille:
Superseeded by https://github.com/fruux/sabre-vobject/pull/379

PR #378, opened 2017-05-22
Fix send message to all users if user is sharee with write access

tbille:
Add argument to parseEvent function in broker: all users with write access on a calendar. This allows to notify the attendees added by someone else that the organizer of an event.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -162,11 +162,16 @@ class Broker {
      *
      * @param VCalendar|string $calendar
      * @param string|array $userHref
+     * @param array $usersWriteAccess
      * @param VCalendar|string $oldCalendar
      *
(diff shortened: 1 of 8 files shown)


evert:
What is the exact problem you're trying to solve?

tbille:
This cases solves te problem for shared calendars and attendees. When a sharee wants to add an attendee to an event that has been created by a another user in a shared calendar, the attendee isn't notified by the scheduler and the event isn't created by it.

This is why instead of notifying only by controlling with the organizer of the event, we want to get all the users that have a write right on the calendar (the organizer but alse all the people that have a write access to the calendar)

DeepDiver1975:
I can hardly judge on this .... will close as it anyhow is conflicting and would need a fresh restart

PR #381, opened 2017-07-03
Fix obvious typo in SCHEDULE-FORCE-SEND handling

alecpl:
Everywhere you use 'forceSend' not 'force-send'.


staabm:
Thx for the fix and sorry for getting back to this that late.

We didnt got notified as you just added a commit but didnt mention anyone.

Thx

PR #383, opened 2017-07-08
Fix possible infinite loop in RRuleIterator.php

mvdnes:
When the RRule FREQ is YEARLY and it uses BYYEARDAY only, an infinite loop could occur.
$this->currentDate may be an DateTimeImmutable, so a reassignment is necessary.


staabm:
Could you provide a unit test which fails without the patch?

mvdnes:
I have added a test.

DominikTo:
Without the change, the new test indeed runs into an infinite loop.

mvdnes:
Is there any timeline on when this will be merged?

staabm:
@mvdnes
Did you run into this problem because a particular client issued a request like that?

Or did you find the problem by studying the codebase and thinking about possible edge cases?

staabm:
@ any maintainer: I dont have enough domain knowledge about this context.

Anyone more in this topic then me: feel free to merge in case this is issue worth beeing fixed

mvdnes:
I think this particular event was imported from an export of my Google calendar. I found the fix by studying (and modifying) the codebase.

staabm:
@BlacklkeEagle can you confirm this fix works?

staabm:
Do we need the same fix for all other occurences of setDate() etc. On date objects in this class?

BlackIkeEagle:
@staabm Yes I can confirm this fix works

BlackIkeEagle:
@staabm In this file this is the only occurence of setDate, but in other files where you want to be able to use DateTime and DateTimeImmutable in the same way this change should be applied to avoid future issues

BlackIkeEagle:
@staabm every setter used on a DateTime or DateTimeImmutable should be converted in best case to the format of $date = $date->set... to allow DateTime and DateTimeImmutable cross use

BlackIkeEagle:
Could this be merged please, this causes "real life" issues

laoneo:
Actually tested it and it worked. Please merge and ship it. @mvdnes, to get it merged, you probably need to update your branch as github tells me that the branch is out of date.

laoneo:
Thanks!

mvdnes:
Thank you for merging!

staabm:
released 4.1.4
https://github.com/sabre-io/vobject/releases/tag/4.1.4

laoneo:
Nice Christmas present :smile:

BlackIkeEagle:
Thanks for merging this fix

Issue #387, opened 2017-09-27
Release Request


lars-sh:
Sorry for the request, but might you finish another minor release in the near future? I know, not that much changed since December 2016, but Request #384 fixes an ugly bug in ownCloud and Nextcloud (https://github.com/nextcloud/contacts/issues/206) and some of us'd like to see that fix being applied there, too. Thanks and much appreciation.

staabm:
In case you PR a proper changelog I might find time to merge and tag something

PR #388, opened 2017-10-17
Updated changelog, bumped version

lars-sh:
As discussed at #387

### DIFF
--- a/lib/Version.php
+++ b/lib/Version.php
@@ -14,6 +14,6 @@ class Version {
     /**
      * Full version number.
      */
-    const VERSION = '4.1.4';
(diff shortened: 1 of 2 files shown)


lib/Version.php
```diff
@@ -14,6 +14,6 @@ class Version {
     /**
      * Full version number.
      */
-    const VERSION = '4.1.3';
+    const VERSION = '4.1.4';
```
staabm: 4.1.3 is correct

lars-sh: Updated pull request

staabm:
thx

staabm:
4.1.3. released

PR #392, opened 2017-10-19
Improve significant change detection in RRULE to limit false positives

alecpl:
Fixes part of #126.

Ignores insignificant changes in RRULE. E.g. ignores INTERVAL=1 and order of items.


evert:
This is really great!

PR #396, opened 2017-11-13
avoid deadlock in nextYearly

BlackIkeEagle:
the while true in byWeekNo or byYearDay can result in a deadlock, when
you don't find the new date within 2000 iterations fallback to the
'easiest form'

Signed-off-by: BlackEagle <ike.devolder@gmail.com>


BlackIkeEagle:
Closed as a duplicate of #383 which is a more proper fix

PR #391, opened 2017-10-18
Support case-insensitive CHARSET Property for vCard 2.1

PHPGangsta:
fixes #390

Issue #390, opened 2017-10-18
Support case-insensitive CHARSET Property for vCard 2.1


PHPGangsta:
For history, see #288 

Yahoo for example has the following vcf file (redacted+reduced):
```
BEGIN:VCARD
VERSION:2.1
SOURCE:Yahoo Contacts (http://contacts.yahoo.com)
PRODID:-//Yahoo/Contacts/YAB3/2017
FN;CHARSET=utf-8:Test
N;CHARSET=utf-8:;Test2;;;
EMAIL;INTERNET;PREF;CHARSET=utf-8:test@test.de
UID;CHARSET=utf-8:07371dc9262a3eb457910d4375d8aa58
X-CID;CHARSET=utf-8:200
X-CREATED;CHARSET=utf-8:1454497086
X-MODIFIED;CHARSET=utf-8:1493525644
X-PRIMARY-PHONE;CHARSET=utf-8:0
END:VCARD
```

I'm getting
```Uncaught Sabre\VObject\ParseException: Unsupported CHARSET: utf-8```

A strtolower() in the switch-condition, and changing all case-values to lowercase, would fix the problem I guess:
https://github.com/sabre-io/vobject/blob/master/lib/Parser/MimeDir.php#L467

Will try to add a Pull Request soon :-)

PR #398, opened 2017-12-06
Allow utf-8 as charset

rullzer:
Even if not strictly valid. Outlook sometimes sets te charset to utf-8
(lowercase) on certain properties. So better just accept this.


PHPGangsta:
Duplicate of #391, see also issue #390

rullzer:
Ah my bad.

PR #399, opened 2017-12-22
Modify the date to check in yearly test correctly as it is immutable

laoneo:
As the current date is immutable in the RRuleIterator, the yearly date needs to be reassigned in the check, otherwise it will be always the same date and turn out in an infinite loop.


laoneo:
Duplicate of #383. Damn wasted a whole morning.

PR #404, opened 2018-03-07
Serialization: Performance boost for long properties

gharlan:
This is a comparison of a CardDAV call in my application, requesting 22 contacts:

screenshot 2018-03-07 12 47 12

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -249,8 +249,9 @@ abstract class Property extends Node {
         $out = '';
         while (strlen($str) > 0) {
             if (strlen($str) > 75) {
-                $out .= mb_strcut($str, 0, 75, 'utf-8') . "\r\n";
-                $str = ' ' . mb_strcut($str, 75, strlen($str), 'utf-8');
+                $part = mb_strcut($str, 0, 75, 'utf-8');
+                $out .= $part . "\r\n";
+                $str = ' ' . substr($str, strlen($part));
             } else {
                 $out .= $str . "\r\n";
                 $str = '';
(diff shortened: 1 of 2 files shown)


lib/Property.php
```diff
@@ -249,8 +249,9 @@ function serialize() {
         $out = '';
         while (strlen($str) > 0) {
             if (strlen($str) > 75) {
-                $out .= mb_strcut($str, 0, 75, 'utf-8') . "\r\n";
-                $str = ' ' . mb_strcut($str, 75, strlen($str), 'utf-8');
+                $part = mb_strcut($str, 0, 75, 'utf-8');
+                $out .= $part . "\r\n";
+                $str = ' ' . substr($str, strlen($part));
```
staabm: could this now scrumble a multibyte char?

staabm: will `strlen($part)` count the correct number of characters when `$part` contains multibyte chars?

staabm: I guess we should check whether we have unit tests for those multibyte cases...

gharlan: strlen counts the bytes, not characters. but together with `substr` it will work correctly.
the multibyte-safe cut position is still determined by `mb_strcut`.

gharlan: There is one test case for this: https://github.com/sabre-io/vobject/blob/master/tests/VObject/PropertyTest.php#L161-L169

gharlan: The test case has more then 75 chars.
"PROPNAME:" (9) + 65 x "!" + "\xc3\xa4bla".
So, the multibyte character is on position 75 + 76.

staabm: you are right, thx.

staabm:
will merge it in case @DeepDiver1975 how now further questions/problems with it.

PR #408, opened 2018-04-18
use full qualified function calls in Property/Text->serialize()

staabm:
explcitly use full qualified function calls so PHP can generate faster opcodes and doesn't need to check the local namespace for equal named functions before using the native function.

applied the changes only in the HOT path.

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -247,11 +247,11 @@ abstract class Property extends Node {
         $str .= ':' . $this->getRawMimeDirValue();
 
         $out = '';
-        while (($len = strlen($str)) > 0) {
+        while (($len = \strlen($str)) > 0) {
             if ($len > 75) {
-                $part = mb_strcut($str, 0, 75, 'utf-8');
+                $part = \mb_strcut($str, 0, 75, 'utf-8');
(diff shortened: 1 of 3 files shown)


staabm:
@gharlan could you double check whether this change improves performance for you?

lib/Property.php
```diff
@@ -247,11 +247,11 @@ function serialize() {
         $str .= ':' . $this->getRawMimeDirValue();
 
         $out = '';
-        while (($len = strlen($str)) > 0) {
+        while (($len = \strlen($str)) > 0) {
             if ($len > 75) {
```
staabm: @evert where does this 75byte limitation come from? is this defined in the spec?

gharlan: @staabm for vcard you can find it here: https://tools.ietf.org/html/rfc6350#section-3.2

staabm:
with this changes php seems to be able to eliminate all the `strlen()` calls (I guess via opcache optimizations). In our case this means ~420.000 less function invocations.

staabm:
including the last commit 3086b8f we get another 2 seconds back.

grafik

PR #412, opened 2018-05-10
Broker: add timezone to CANCEL messages

bosim:
If DTSTART and/or DTEND and other times are specified with a TZID, some clients
may fail reading the generated iCal data, since the VTIMEZONE definition is
missing.


DeepDiver1975:
I'll backport this to 3.2 branch and prepare a release. This is breaking cancellation of events with clients using Exchange. (and produces invalid ics attachments ;-) )

DeepDiver1975:
> I'll backport this to 4.1 branch and prepare a release. This is breaking cancellation of events with clients using Exchange. (and produces invalid ics attachments ;-) )

maybe not needed .....

PR #415, opened 2018-06-24
Fixed poor implementations of vCard 3.0 being rejected by server

Webbeh:
Let's take an example of the BlackBerry 10 OS.

It shares the vCards in the v3 format, but uses the "BASE64" encoding instead of the "B" one. This results in really odd behaviour in the communication between the sabre server and the BB client  : 

- contact disappearing from both the server and client
- contact being duplicated in the client at each and every synchronisation
- contact losing its name, phone number and/or profile picture
- ...

This PR effectively fixes that issue for each and every OS10 based BlackBerry device, and I assume will fix every client implementation that uses the "BASE64" encoding instead of "B".

I have tested this patch successfully in both sabre standalone and in Nextcloud 12 and 13.

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -616,7 +616,7 @@ abstract class Property extends Node {
                         $allowedEncoding = ['QUOTED-PRINTABLE', 'BASE64', '8BIT'];
                         break;
                     case Document::VCARD30 :
-                        $allowedEncoding = ['B'];
+                        $allowedEncoding = ['B', 'BASE64'];
                         break;
 
                 }


evert:
`ENCODING=BASE64` is not valid in vCard 3. Supporting broken vCard implementations is kind of a bad idea, because it means that sabre/dav will also serve broken vCards to other clients.

I think you should report the bug first to blackberry and see if they can fix it. If they refuse or don't respond, we can reopen this discussion.

Webbeh:
BlackBerry dropped support for OS10 last year.

Any chance in accepting the reception of BASE64 vCards and doing the stripping of the extra 5 chars on the fly?

evert:
I didn't know!

In that case, I would propose the following:

Change your PR to not remove the error, but instead 'repair' the problem (with the `REPAIR` option) and emit a warning instead of an error. Repairing in this case is simply changing the `BASE64` into `B`

sabre/dav will call validate with the repair option by default, so this should fix your problem in a more correct way

Webbeh:
Will try that tonight.

Webbeh:
I see this has been closed.

Has a fix been made available anywhere ? Or an alternative ?

msberends:
Please reopen; there is no fix yet! `ENCODING=BASE64` is still not supported! @evert

evert:
@msberends if you want to provide a good PR for this, it would get accepted.

Webbeh:
image
...Would this work ?

Issue #418, opened 2018-09-12
VTIMEZONE transitions in calendar feed generation


jpirkey:
Am I missing something or is there not support for having the calendar generation include VTIMEZONE entries for DST and Standard transitions?

I found [this](https://gist.github.com/thomascube/47ff7d530244c669825736b10877a200), but I wouldn't do it this way.  I would just use `RDATE` entries per offset and transition.

Something like this:
```ics
BEGIN:VTIMEZONE
TZID:America/New_York
X-MICROSOFT-CDO-TZID:10
BEGIN:DAYLIGHT
DTSTART:19700426T070000Z
TZOFFSETFROM:-500
TZOFFSETTO:-400
TZNAME:EDT
RDATE:20080309T070000
RDATE:20090308T070000
RDATE:20100314T070000
RDATE:20110313T070000
RDATE:20120311T070000
RDATE:20130310T070000
RDATE:20140309T070000
RDATE:20150308T070000
RDATE:20160313T070000
RDATE:20170312T070000
RDATE:20180311T070000
RDATE:20190310T070000
RDATE:20200308T070000
RDATE:20210314T070000
RDATE:20220313T070000
RDATE:20230312T070000
RDATE:20240310T070000
RDATE:20250309T070000
RDATE:20260308T070000
RDATE:20270314T070000
RDATE:20280312T070000
END:DAYLIGHT
BEGIN:STANDARD
DTSTART:19701025T060000Z
TZOFFSETFROM:-400
TZOFFSETTO:-500
TZNAME:EST
RDATE:20081102T060000
RDATE:20091101T060000
RDATE:20101107T060000
RDATE:20111106T060000
RDATE:20121104T060000
RDATE:20131103T060000
RDATE:20141102T060000
RDATE:20151101T060000
RDATE:20161106T060000
RDATE:20171105T060000
RDATE:20181104T060000
RDATE:20191103T060000
RDATE:20201101T060000
RDATE:20211107T060000
RDATE:20221106T060000
RDATE:20231105T060000
RDATE:20241103T060000
RDATE:20251102T060000
RDATE:20261101T060000
RDATE:20271107T060000
END:STANDARD
END:VTIMEZONE
```

And I was thinking of adding it to `VCalendar` object to add support for generating the `VTIMEZONE` entries based on the events in the calendar via a method call (something like `generateVTimezoneEntries`) that could be called before the calendar's serialize method.

Thoughts?

evert:
[...]

PR #419, opened 2018-09-12
Additional vtimezone validation rules

jpirkey:
see also #418 

Reference: https://www.kanzaki.com/docs/ical/vtimezone.html

### DIFF
--- a/lib/Component/VTimeZone.php
+++ b/lib/Component/VTimeZone.php
@@ -49,9 +49,17 @@ class VTimeZone extends VObject\Component {
 
         return [
             'TZID' => 1,
+            'DTSTART' => 1,
+            'TZOFFSETTO' => 1,
+            'TZOFFSETFROM' => 1,
 
             'LAST-MODIFIED' => '?',
-            'TZURL'         => '?',
+            'TZURL' => '?',
+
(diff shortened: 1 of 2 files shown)


jpirkey:
Two tests still fail, but I'd like some input as to if that is truly expected behavior and we need to support that or not:
image

jpirkey:
@evert can you give me an idea who the new maintainers are?  They seem to be less responsive then you ;-)

renaudboyer:
ICal test is failing because the following rules cannot work with VTimezone validation rules: 
```
'DTSTART' => 1,
'TZOFFSETTO' => 1,
'TZOFFSETFROM' => 1,
```
They are are correct within `DAYLIGHT` and `STANDARD` but not in `TIMEZONE` as `DAYLIGHT` and `STANDARD` are `0+`
With the way VTimezone work, rules should be:
```
'DTSTART' => '*',
'TZOFFSETTO' => '*',
'TZOFFSETFROM' => '*',
```

jpirkey:
Riiiight.  I see that now.  Thanks!

renaudboyer:
I Agree and as '*' means no verification, I think we should not add them.

Issue #422, opened 2018-09-21
PHP 7.3 compatibility


alecpl:
First problem PHP Parse error:  syntax error, unexpected ':' in /home/alec/repos/vobject/tests/VObject/BirthdayCalendarGeneratorTest.php on line 480

And another one (after I fixed the first):
```
1) Sabre\VObject\BirthdayCalendarGeneratorTest::testVcardStringWithValidBirthday
"continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?

/home/alec/repos/vobject/lib/Component/VCard.php:514
/home/alec/repos/vobject/lib/Parser/MimeDir.php:166
/home/alec/repos/vobject/lib/Parser/MimeDir.php:89
/home/alec/repos/vobject/lib/Reader.php:46
/home/alec/repos/vobject/lib/BirthdayCalendarGenerator.php:73
/home/alec/repos/vobject/tests/VObject/BirthdayCalendarGeneratorTest.php:39

2) Sabre\VObject\FreeBusyGeneratorTest::testGeneratorBaseObject
"continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?

/home/alec/repos/vobject/lib/FreeBusyGenerator.php:433
/home/alec/repos/vobject/tests/VObject/FreeBusyGeneratorTest.php:16
```

alecpl:
Merged.

PR #423, opened 2018-09-21
Fix PHP 7.3 compatibility (#422)

alecpl:


alecpl:
Indeed it looks that in these cases just `break` is fine. Changed.

DeepDiver1975:
assimilated into https://github.com/sabre-io/vobject/pull/424 - Thanks a lot for your contribution!

PR #424, opened 2018-09-24
Test with php 7.3

DeepDiver1975:


DeepDiver1975:
@staabm seems like there is no way to test with the current travis config on php 5.5

I vote for removing php 5.5 from travis for now - later we can decide to drop it completly ...

Objections?

evert:
Why don't you cut a new major version and drop PHP < 7? Might make your life a bit easier.

DeepDiver1975:
We can drop older version in a followup pr.

@staabm merge? We can then tag a new alpha and setup dependencies in other repos properly.

staabm:
Thanks

staabm:
4.2.0. when we dont drop php5, agree

DeepDiver1975:
here we go - https://github.com/sabre-io/vobject/releases/tag/4.2.0-alpha1

PR #427, opened 2018-10-05
Use php-cs-fixer 2.x

DeepDiver1975:
.php_cs.dist holds a converted config of the rules as specified in https://github.com/sabre-io/cs/blob/master/bin/php-cs-fixer-config.php

Looks like php-cs-fixer 2.x can fix more double quotes to single quotes cases ...


.travis.yml
```diff
@@ -15,19 +15,24 @@ addons:
       # required for php7.3
       - libzip4
 
-script:
-  - ./bin/phpunit --configuration tests/phpunit.xml --coverage-clover=coverage.xml
-  - ./bin/sabre-cs-fixer fix . --dry-run --diff
+install:
+  - mkdir -p tools
+  - wget -nc -P tools https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar
```
staabm: I would prefer prettyci instead of a travis build formthe style check
https://prettyci.com/

That way the travis builds dont last even longer

DeepDiver1975: cool - I'll have a look. I was not aware of this service. THX

staabm: I just added prettyci to all sabre repositories.

We only need the nen .trvais.yml changes for this PR then

.php_cs.dist
```diff
@@ -0,0 +1,63 @@
+<?php
+
+$config = PhpCsFixer\Config::create();
+$config->getFinder()
+    ->exclude('vendor')
+    ->in(__DIR__);
+$config->setRules([
+    '@PSR1' => true,
```
staabm: I dont mind using a defualt style like psr2 or symfony-style or similar.

I dont feel its worth having a „custom“ style... @evert any opinion on this?

.travis.yml
DeepDiver1975: Reverted all this ....

staabm:
it seems the check doesnt work right now.. I contacted the prettyci author 

https://twitter.com/markusstaab/status/1049557466609045504

PR #432, opened 2018-10-25
DTSTAMP must be specified in UTC

Tragen:
DTSTAMP must be specified in UTC
See https://www.kanzaki.com/docs/ical/dtstamp.html

### DIFF
--- a/lib/Component/VEvent.php
+++ b/lib/Component/VEvent.php
@@ -83,7 +83,7 @@ class VEvent extends VObject\Component
     {
         return [
             'UID' => 'sabre-vobject-'.VObject\UUIDUtil::getUUID(),
-            'DTSTAMP' => date('Ymd\\THis\\Z'),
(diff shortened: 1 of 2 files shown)


staabm:
I am wondering, that we dont need to adjust a testcase. I guess this method is not covered.

@Tragen would you mind adding a testcase? dont worry if you dont feel confident enought to do it though.

DeepDiver1975:
> I am wondering, that we dont need to adjust a testcase. I guess this method is not covered.

php env has timezone set to utc? In this case this change would have no impact - right?

evert:
You can't assume the timezone is set to UTC though. It's highly recommended but often not true.

DeepDiver1975:
> You can't assume the timezone is set to UTC though.

sure - my comment was about the fact that tests are passing with this change

staabm:
Ok, no problem. Thx for the PR though

PR #434, opened 2018-10-29
Update Broker.php, to force the test of ORGANIZER e-mail address to b…

johnkwoods:
…e case-insensitive

Fixes issue where a recurring meeting is created with ORGANIZER as First.Last@domain.com, then individual appointments are saved with ORGANIZER as first.last@domain.com.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -847,7 +847,7 @@ class Broker
                     $organizer = $vevent->ORGANIZER->getNormalizedValue();
                     $organizerName = isset($vevent->ORGANIZER['CN']) ? $vevent->ORGANIZER['CN'] : null;
                 } else {
-                    if ($organizer !== $vevent->ORGANIZER->getNormalizedValue()) {
+                    if (strtolower($organizer) !== strtolower($vevent->ORGANIZER->getNormalizedValue())) {
                         throw new SameOrganizerForAllComponentsException('Every instance of the event must have the same organizer.');
                     }
                 }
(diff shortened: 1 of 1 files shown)


lib/ITip/Broker.php
```diff
@@ -847,7 +847,7 @@ protected function parseEventInfo(VCalendar $calendar = null)
                     $organizer = $vevent->ORGANIZER->getNormalizedValue();
                     $organizerName = isset($vevent->ORGANIZER['CN']) ? $vevent->ORGANIZER['CN'] : null;
                 } else {
-                    if ($organizer !== $vevent->ORGANIZER->getNormalizedValue()) {
+                    if (strtolower($organizer) !== strtolower($vevent->ORGANIZER->getNormalizedValue())) {
```
staabm: Logically doesnt make a difference, but I would use `strtoupper` as we use in the below lines

staabm:
Would you mind adding a unit test?

lib/ITip/Broker.php
johnkwoods: Feel free to change to strtoupper, if you would like.

johnkwoods:
> Thx for your PR!
> 
> Would you mind adding a unit test?

You're welcome.

Unfortunately, I'm a fish-out-of-water when it comes to writing tests... Might be better served with someone who actually knows what they're doing there.

lib/ITip/Broker.php
staabm:
```suggestion
                    if (strtoupper($organizer) !== strtoupper($vevent->ORGANIZER->getNormalizedValue())) {
```

DeepDiver1975:
see #435

PR #435, opened 2018-10-30
Johnkwoods patch 1

DeepDiver1975:
#434 with unit tests


staabm:
thx @DeepDiver1975 <3

PR #436, opened 2018-11-28
Fix bug in by year day

KAYLukas:
The `BYYEARDAY` parameter is interpreted incorrectly, the day seems to be off by 1. `BYYEARDAY=1` indicates currently the second day of the year. Seems to be incorrect according to the RFC. Also double checked this against iCal, which confirmed my suspicion.

Furthermore, the negative part also seems to be handled incorrectly.


valentinbonneaud:
This bug is still present on `master`. 

You can observe the issue by generating occurences using the following rule: `FREQ=YEARLY;COUNT=3;BYYEARDAY=1` and the following start date `2011-01-01 03:07:00`. The currently generated times will be 

```
[
    '2011-01-01 03:07:00',
    '2011-01-02 03:07:00',
    '2012-01-02 03:07:00',
]
```
instead of
```
[
    '2011-01-01 03:07:00',
    '2012-01-01 03:07:00',
    '2013-01-01 03:07:00',
]
```

> The BYYEARDAY rule part specifies a COMMA-separated list of days of the year.  Valid values are 1 to 366 or -366 to -1.  For example, -1 represents the last day of the year (December 31st) and -306 represents the 306th to the last day of the year (March 1st).  The BYYEARDAY rule part MUST NOT be specified when the FREQ rule part is set to DAILY, WEEKLY, or MONTHLY.

https://tools.ietf.org/html/rfc5545#section-3.3.10

Based on the RFC, `BYYEARDAY=1` should return the first day of the year and not the second.

phil-davis (approved):

phil-davis:
@KAYLukas I rebased this in #582 and added a few more test cases. It is a real bug. Thanks for the contribution. I have merged it in #582

PR #438, opened 2018-12-07
Enhance fastforward speed if no `count` value has been given

KAYLukas:


staabm:
Thanks for the PR.

This adds a lot of code. For which use-case do these changes improve performance? Did you measure how much it improves performance?

KAYLukas:
@staabm It's dependent on the other PRs (which are in the same branch) because otherwise I would have to write tests that are on purpose incorrect. 

It improves performance for basically every rrule that does not have the `COUNT` limit that is included. I'm working on a good benchmark, I did benchmark it a few weeks ago, just lost the data ;-)

evert:
I'm also a little worried about the complexity this adds. It's not a super easy to maintain part of the code. At the very least the newly introduced code is not really documented, and it could stand to benefit from more in-line documentation and better variable naming. It's pretty impressive though. I have no doubt that this greatly improves performance.

staabm:
Yeah its a lot of complexity. I cant overlook the impact of the change though, therefore will delegate the final call to @evert

evert:
For what it's worth, the intent of the `fastForward` function (for me) was always to have a hook where future optimizations (exactly such as these) can happen. 

The way I envisioned that these optimizations would happen, is that the fastForward function looks if the rules from RRULE are considered "simple enough", such as `RRULE:FREQ=DAILY`, and instead of calling `nextDaily()` just immediately skip ahead to those dates.

By giving every `nextXXX()` function an `amount` argument, the fast-forwarding logic is effectively handed off to those respective functions, except those functions themselves don't respect `COUNT` or `UNTIL`. Thankfully they're `protected`.

So usually what I will ask people who need a feature like this is: Are there 2 or 3 specific cases that would solve 90% of your problem, because that tends to be a bit more effective in the long. For example, having support for faster fastForwarding in hourly events might not be interesting to anyone because hourly recurring events are so rare.

I think another feeling with this patch is that it feels very much like a patch on existing logic, when maybe the existing logic/structure wasn't really good enough. This makes me wonder if a better design exists to satisfy this requirement.

Anyway, I realize that these are fairly vague and not-that-constructive comments :/

PR #441, opened 2018-12-24
Repairing BASE64 encoded vCard version 3

Webbeh:
Follow up of
#415

Repairs vCards in V3 that have a BASE64 encoding.

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -574,9 +574,9 @@ abstract class Property extends Node
                         $allowedEncoding = ['B'];
                         //Repair vCard30 that use BASE64 encoding
                         if ($options & self::REPAIR) {
-                            if ('BASE64' == strtoupper($encoding)) {
+                            if ('BASE64' === strtoupper($encoding)) {
                                 $encoding = 'B';
-                                $this->offsetSet('ENCODING', $encoding);
+                                $this['ENCODING'] = $encoding;
                                 $warnings[] = [
                                     'level' => 1,
                                     'message' => 'ENCODING=BASE64 has been transformed to ENCODING=B.',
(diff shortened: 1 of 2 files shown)


Webbeh:
What's the point of the "offsetSet" function if you're not using it ?

Also the identity operator shouldn't be necessary in this case, as strtoupper() always returns a string, but I guess it doesn't hurt.

evert:
> What's the point of the "offsetSet" function if you're not using it ? Just wondering.

The `offsetSet` function is part of the ArrayAccess PHP interface. This interface's sole role is to make sure that `$obj['item']` can be used.

> Also the identity operator shouldn't be necessary in this case, as strtoupper() always returns a string, but I guess it doesn't hurt.

Apparently it's slightly faster, but it's probably negligible. As a general rule I find that using string comparisons (`===` and `!==`) should be used always, _except_ if you deliberately want to opt-out of strict checking and must use `==` for specific situations. Applying this rule strictly can avoid subtle bugs.

In this case neither of my suggestions actually did anything, it's just small stuff to help guard consistency. I also never used the 'suggest' feature before so I hoped you could just accept the suggestion with the click of a button ;)

Webbeh:
It was not just "one click" but close enough :P 

Did you have to override the offsetSet method ? I don't see what more it could do compared to standard arrays.

Yeah I know, it was just for consistency. It makes sense than the identity operator is slightly faster, since it doesn't have to bother with type casting. However, I'm assuming that if you provide two identical objects (let's say two strings with the value of "foo"), it still has to check whether the type is the same in both cases : in this one to check whether it's actually the same type or not, and with the equality operator to check whether to cast the object or not.

PR #443, opened 2019-01-08
we stop before the year 10000

### LINKED ISSUE #442: Never reoccurring recurrence loops forever
Consider this recurrence rule:

`RRULE:FREQ=MONTHLY;BYDAY=2TU;BYSETPOS=2`

Systems like google accept this as a valid rule, but it never reoccurs... There is only one "second Tuesday" in each month. Putting this into a Google event results in a recurring event that occurs on the first date and never again.

The `RRuleIterator` becomes stuck in the `while(true)` loop, looping until the year `9999` to try to find the next occurrence and finally breaking when it hits the year `10000` and the `DateTimeImmutable` constructor fails.

DeepDiver1975:
Fixed #442 

Just before the RRuleIterator hits the year 10.000 we stop iterating.

### DIFF
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -67,6 +67,9 @@ class RRuleIterator implements Iterator
      */
     public function valid()
     {
+        if (null === $this->currentDate) {
+            return false;
+        }
         if (!is_null($this->count)) {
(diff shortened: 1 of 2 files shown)


staabm:
Thx!

PR #447, opened 2019-03-04
Update VCalendar.php

ger82:
The TimeZone was not adopted correctly and was always replaced with UTC.

### DIFF
--- a/lib/Component/VCalendar.php
+++ b/lib/Component/VCalendar.php
@@ -299,7 +299,7 @@ class VCalendar extends VObject\Document
                     // We only need to update the first timezone, because
                     // setDateTimes will match all other timezones to the
                     // first.
-                    $dt[0] = $dt[0]->setTimeZone(new DateTimeZone('UTC'));
+                    $dt[0] = $dt[0]->setTimeZone($timeZone);
                     $componentChild->setDateTimes($dt);
                 } elseif ($componentChild instanceof Component) {
                     $stripTimezones($componentChild);


DeepDiver1975:
Unit tests are failing because if this change. I can hardly judge on this change .....

staabm:
I cannot say whats correct either.

evert:
It does convert to UTC by design. The documentation even mentions this as the intended behavior.

CalDAV relies on this for the calendar query requests. If sombody would want a version of this function that doesn't convert timezones, I can imagine this could be offered as an optional parameter, but it would need to be done in a non-bc breaking way and then with tests

ger82:
where can I find this documentation?

evert:
Hi @ger82,

That's the documentation I was talking about. Just so this is super clear, concerting to UTC is *by design* and changing this will break CalDAV clients

ger82:
Hello evert,

so I think I do something wrong. I wont to use my timezone.
How can I do this?
This ist my code:

`include 'vobject/vendor/autoload.php';
use
	Sabre\VObject;

$calData = [];
$calendarAdress = ['icalendartest.ics'];

foreach($calendarAdress as $adress) {
	$calendar = VObject\Reader::read( file_get_contents($adress) );

	$today = new DateTime();
	$today->setTime(00, 00);
	$nextWeek = new DateTime('+7 days');
	$newVCalendar = $calendar->expand( $today, $nextWeek, new DateTimeZone('Europe/Berlin') );

	if(!empty($newVCalendar->vevent)){
		foreach($newVCalendar->vevent as $event) {
			$dtstart = new DateTime( (string)$event->dtstart );
			
			echo	(string)$dtstart->format('d-m-Y H:i'), " ",$dtend->format('d-m-Y H:i'), " ", (string)$event->summary ,"<br>";
		}
	}
}`

evert:
There is nothing wrong with wanting this feature but it has to be built in a way that doesn't break backwards compatibility and with tests

ger82:
Thank you evert,

can you give me a hint how I can do this for my example.

evert:
If you need this to be explained, just open up a regular feature request instead.

PR #450, opened 2019-03-17
Fix parse with option Forgiving with trailing equal

asbiin:
When using option Forgiving, if a quoted printable line ends with an equal sign, followed by en empty line, there will be a parse error.
In this case, the `readLine` must not be called again.


staabm:
Thank you for the PR. Could you rebase it? It got stale after mergine other PRs

staabm:
Thx

PR #461, opened 2019-06-24
fix phpstan level1 warnings

staabm:
after we fixed these warnings php IDEs can get smarter about the classes runtime properties and also other static analysis tools can make better assumptions/decisions.

primary goal of this PR is fixing the phpstan level1 warnings (therefore a green build), not defining all possible properties. I dont have enough knowledge to do this.

### DIFF
--- a/lib/Component/Available.php
+++ b/lib/Component/Available.php
@@ -14,9 +14,9 @@ use Sabre\VObject;
  * @author Ivan Enderlin
  * @license http://sabre.io/license/ Modified BSD License
  *
- * @property \DateTime $DTSTART
+ * @property \DateTime      $DTSTART
  * @property \DateTime|null $DTEND
- * @property mixed $DURATION
+ * @property mixed          $DURATION
  */
 class Available extends VObject\Component
 {
--- a/lib/Component/Available.php
+++ b/lib/Component/Available.php
@@ -13,6 +13,9 @@ use Sabre\VObject;
  * @copyright Copyright (C) fruux GmbH (https://fruux.com/)
  * @author Ivan Enderlin
  * @license http://sabre.io/license/ Modified BSD License
+ *
+ * @property $DTSTART
+ * @property $DURATION
  */
 class Available extends VObject\Component
 {
--- a/lib/Component/Available.php
+++ b/lib/Component/Available.php
@@ -14,8 +14,9 @@ use Sabre\VObject;
  * @author Ivan Enderlin
  * @license http://sabre.io/license/ Modified BSD License
  *
- * @property $DTSTART
- * @property $DURATION
+ * @property \DateTime $DTSTART
+ * @property \DateTime|null $DTEND
+ * @property mixed $DURATION
  */
 class Available extends VObject\Component
 {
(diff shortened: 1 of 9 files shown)


lib/Component/Available.php
```diff
@@ -13,6 +13,10 @@
  * @copyright Copyright (C) fruux GmbH (https://fruux.com/)
  * @author Ivan Enderlin
  * @license http://sabre.io/license/ Modified BSD License
+ *
+ * @property \DateTime $DTSTART
```
staabm: I dont know very much about how this lib works, therefore reverse engineering the types based on the implmentation.

in case someone knows more then me and can help me to make the possible types stricter, feel free to jump in @evert @DominikTo @Hywan

evert: Is it possible to make this something like typescript's `any` ?

staabm: not sure what `any` is in TS. but we can used `mixed` in case you want to allow everything.

staabm: but our implementation of this class requires at least a object which support a `getDateTime()` method. everything else would create a runtime error in `getEffectiveStartEnd()`

evert: `mixed` sounds like the sane thing. I think having support for any property that can be any type might make this a lot simpler. Being pretty grounded into the Javasript / Typescript ecosystem now I don't think this style of Sabre\VObject design works well with static type analyzers. It's hard to be comprehensive because types can also be overridden with the `TYPE` parameter.

staabm: having the properties defined as `mixed` would at least make the tools aware that those properties exist.

I dont get why mixed would be usefull here because of `getEffectiveStartEnd()` only working with a `getDateTime()` method though.
I expected it expects a DateTime object - but was just a guess

evert: It can almost expect any type. It's not dependent on the property name

staabm: thats true for the base class `VObject\Component` but for the concrete subclasses we have a more concrete set of properties and values, no?

evert: As far as I know many properties can have a multitude of types, and even more types are accepted when setting (as opposed to getting) properties.

evert: The real issue I have is that the types are just very incomplete, which means that anyone using this library with phpstan or other phpdoc tools are going to get a barrage of errors even when they're using the library as intended.

So the types need to be either strict and exhaustive or not there at all.

staabm: Hmm ok. So will change everything to mixed then

evert: I think that's the safest. Ideally also just assume that any property can exist =)

JeroenVanOort:
Wouldn't want one to revert https://github.com/sabre-io/vobject/commit/ebe974884ef45cd6b62472b5942f8839ae48b5c9 as part of this PR?

staabm:
I will not work here anymore. Feel free to takeover

staabm:
Superseded by https://github.com/sabre-io/vobject/pull/487

PR #463, opened 2019-07-03
Issue 462 - Broker::parseEventForOrganizer copies DTSTAMP from $eventInfo that causes broken scheduling

alexander-ivannikov:
Implementation of a fix for https://github.com/sabre-io/vobject/issues/462


staabm:
It seems this requires some changes to test-expectations.

staabm:
you are on the right track. just one unit test failure left :-)

alexander-ivannikov:
> you are on the right track. just one unit test failure left :-)

Looks like I have finally fixed all of the failing unit tests. But not the entire CI build...

tests/VObject/ITip/EvolutionTest.php
```diff
@@ -1719,6 +1717,7 @@ public function testNewEvolutionEvent()
  TRUE;LANGUAGE=en:MAILTO:dominik@fruux.com
 CREATED:20140813T153211Z
 LAST-MODIFIED:20140813T155353Z
+DTSTAMP:**ANY**
```
vimmerru: @alexander-ivannikov We can check that DTSTAMP was increased and it is more than moment before parseEventForOrganizer call

staabm:
thx - we will do the phpstan in a separate PR.

PR #465, opened 2019-07-11
Add TZ in iTip REPLY iTip messages

renaudboyer:
Add VTIMEZONE objects in iTip REPLY if there were present in the original message.

Allows to have well-formatted iTip REPLY messages


Julian1998:
@DeepDiver1975 Can you merge this one?

DeepDiver1975:
Looks good to me - I'd like to have a second pair of eyes - @staabm maybe? THX

staabm:
@evert should have a look. I am no expert in this area

m-a-v:
I still haven't managed to get correct REPLY for Microsoft Outlook or Google invitations (if sent as attendee). Should this work with this fix?

Julian1998:
@m-a-v Can you validate your ics and tell what happened wrong by using https://icalendar.org/validator.html

Not sure what you mean by getting a correct REPLY. But this is a fix for several clients which fail when rendering an event with a TZID set (and the VTIMEZONE object misses ofc)

renaudboyer:
We were working on Microsoft/Google compatibility issues when we've done this fix. I don't remember which brand was rejecting the reply messages because of invalid format.
If you're using TZ, it should fix your replies.

DeepDiver1975:
> @evert should have a look. I am not an expert in this area

@evert any chance to have a look at this? THX a lot

PR #469, opened 2019-10-09
fix  Bad file descriptor (7.4)

remicollet:
With 7.4.0RC3
```
There was 1 error:

1) Sabre\VObject\Parser\MimeDirTest::testParseError
fgets(): read of 8192 bytes failed with errno=9 Bad file descriptor

```

This minimal fix open file in read/write mode, to avoid the new warning.

A better fix will be to detect if the stream is properly open for read in setInput 
```
 is_resource 
 get_resource_type == "stream"
 stream_get_meta_data => "mode"
``` 
But this will be a new feature, so far away for the goal of this simple fix.

### DIFF
--- a/tests/VObject/Parser/MimeDirTest.php
+++ b/tests/VObject/Parser/MimeDirTest.php
@@ -16,7 +16,7 @@ class MimeDirTest extends TestCase
     public function testParseError()
     {
         $mimeDir = new MimeDir();
-        $mimeDir->parse(fopen(__FILE__, 'a'));
+        $mimeDir->parse(fopen(__FILE__, 'a+'));
     }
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
failing PHPStan is unrelated - :+1: from my pov

THX @remicollet

remicollet:
For memory, in UPGRADINGS

    . fread() and fwrite() on plain files will now throw a notice on failure,
      such as when trying to write to a read-only file resource.

staabm:
thx

staabm:
will allow-failures for phpstan in https://github.com/sabre-io/vobject/pull/470

PR #474, opened 2019-10-14
add optional datetime range filter when parsing large calendar files

### LINKED ISSUE #473: Parsing large calendar files.
I'm having to parse a large VCalendar file with over 2000 VEvents and i was looking for a method to help reduce the size of the resulting ```VObject\Document``` as the resulting object is overkill for what i'm actually using it for. 
Most of the events in the file are in the past and i only need the events from present to around a month in the future.
I was struggling to find any way to do it with the current project. I've ended up adding these few lines to [```VObject\MimeDir::parseDocument```](https://github.com/sabre-io/vobject/blob/ebe974884ef45cd6b62472b5942f8839ae48b5c9/lib/Parser/MimeDir.php#L142)

```
        $loLim = new \DateTime('now');
        $hiLim = new \DateTime('4 weeks');
        while (true) {
            // Reading until we hit END:
            $line = $this->readLine();
            if ('END:' === strtoupper(substr($line, 0, 4))) {
                break;
            }
            $result = $this->parseLine($line);
            if ($result) {
                if ($result instanceof Component\VEvent) {
                    if (
                        $result->DTSTART->getDateTime() > $hiLim ||
                        $result->DTEND->getDateTime() < $loLim 
                    ) {
                        continue;
                    }
                }
                $this->root->add($result);
            }
        }
```
Preferably the ```$hiLim``` and ```$loLim``` values would be passed in as parameters.
Using this it reduced the memory usage in my case from over 30MB to less than 5MB.

This is just an example of what I've done and was wondering if this is functionality that could benefit the project? Or if there's functionality already included that i just cant find.

8633brown:
closes #473. managed to figure out the ```Component\VEvent::isInTimeRange``` method and figured it was the right method for this task.

### DIFF
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -2,6 +2,8 @@
 
 namespace Sabre\VObject\Parser;
 
+use DateTimeImmutable;
+use DateTimeInterface;
 use Sabre\VObject\Component;
 use Sabre\VObject\Component\VCalendar;
 use Sabre\VObject\Component\VCard;
@@ -40,6 +42,20 @@ class MimeDir extends Parser
     protected $root;
 
     /**
+     * Start of range.
+     *
+     * @var DateTimeInterface|null
+     */
+    protected $start;
+
+    /**
+     * End of range.
+     *
+     * @var DateTimeInterface|null
+     */
+    protected $end;
+
+    /**
      * By default all input will be assumed to be UTF-8.
      *
      * However, both iCalendar and vCard might be encoded using different
@@ -75,7 +91,7 @@ class MimeDir extends Parser
      *
      * @return \Sabre\VObject\Document
      */
-    public function parse($input = null, $options = 0)
+    public function parse($input = null, $options = 0, DateTimeInterface $start = null, DateTimeInterface $end = null)
     {
         $this->root = null;
 
@@ -86,6 +102,8 @@ class MimeDir extends Parser
         if (0 !== $options) {
             $this->options = $options;
         }
+
+        $this->setTimeRange($start, $end);
 
         $this->parseDocument();
 
@@ -137,6 +155,20 @@ class MimeDir extends Parser
     }
 
     /**
+     * Sets the time range.
+     *
+     * Any VEvent object falling outside of this time range will be ignored.
+     *
+     * @param DateTimeInterface $start
+     * @param DateTimeInterface $end
+     */
+    public function setTimeRange(DateTimeInterface $start = null, DateTimeInterface $end = null)
+    {
+        $this->start = $start;
+        $this->end = $end;
+    }
+
+    /**
      * Parses an entire document.
      */
     protected function parseDocument()
@@ -173,6 +205,13 @@ class MimeDir extends Parser
             }
             $result = $this->parseLine($line);
             if ($result) {
+                if ($result instanceof Component\VEvent) {
+                    if ($this->start && $this->end) {
+                        if (! $result->isInTimeRange($this->start, $this->end)) {
+                            continue;
+                        }
+                    }
+                }
                 $this->root->add($result);
             }
         }
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -2,7 +2,6 @@
 
 namespace Sabre\VObject\Parser;
 
-use DateTimeImmutable;
 use DateTimeInterface;
 use Sabre\VObject\Component;
 use Sabre\VObject\Component\VCalendar;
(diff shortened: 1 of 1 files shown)


lib/Parser/MimeDir.php
```diff
@@ -75,7 +91,7 @@ class MimeDir extends Parser
      *
      * @return \Sabre\VObject\Document
      */
-    public function parse($input = null, $options = 0)
+    public function parse($input = null, $options = 0, DateTimeInterface $start = null, DateTimeInterface $end = null)
```
staabm: Do you really need the params here or is the public setTimeRange() method sufficient?

8633brown: Yeah I guess theres no real need for them. Think I left them in to keep it somewhat consistant with the freebusy generator. The call to setTimeRange can be removed from the parse method then too. 
I can do this tomorrow at some point.

evert:
Doing this 'right' might be a more difficult than you think, especially when things like recurrence and exceptions come into play. You might need to do 2 passes of the file. The first to create an index of related events, and a second to do the actual filtering.

I'm a little on the fence here. I also feel that the use-case (reducing 30MB memory to 5MB) is not incredibly significant if this is mostly used for one-off purposes. It's definitely a significant number in terms of memory usage, but is this something you're dealing with very often?

8633brown:
Basically what I'm working on is a personalised teacher student booking site. I dont particularly want to go down the route of self hosting a caldav server but that might be the. Ext best option. The plan was to allow teachers (which theres only 2 of at the moment) to pass a public ical link to the server to be parsed to find the freebusy times for when students can create booking. Most of the calendars I'm dealing with are hosted on the apple servers so I'm struggling to find a sensible way to query these with caldav requests. As mentioned one of the calendars has over 2000 events and once parsed with reader::read and everything is put into memory it around 30MB. Once reduced down to the events I'm interested in (I was wrong about 5MB that was also including another calendar) it's closer to 1.5MB. 
So this action is possibly going to have to happen every time a lesson is modified and every Xhours to keep it upto date (when not considering checking the file for changes or checking Etag headers and the likes.)
I'm very open to any new suggestions thrown my way.

evert:
I guess where I'm sort of coming from is that even at the 30MB point, you can still run about >100 in parallel in theory. If it takes 5 seconds to parse one of these larger ones, 1200 this system should be able to handle 100K of these calendar updates per hour. I'm just wondering: is that really your bottleneck?

8633brown:
You're right its probably not a bottleneck at all. I just thought a method like this could've been a quick win. I was assuming the functionality would more or less be the same as the [```FreeBusyGenerator::calculateBusy```](https://github.com/sabre-io/vobject/blob/ebe974884ef45cd6b62472b5942f8839ae48b5c9/lib/FreeBusyGenerator.php#L392-L458) as this is pretty much identical to the ```VEvent::isInTimeRange``` method. I also seen the min and max dateTimes in the settings which i originally thought would do what i wanted. I guess i'm just struggling to see the difference in parsing the VEvent in the ```FreeBusyGenerator``` when i could just parse them out when i read the calendar the first time. 
Apologies if i'm just being naive thinking this is easier than it is.

evert:
@8633brown there is a difference with the FreeBusyIterator, because this object actually right now needs the full calendar in memory to do its thing. It specifically uses `EventIterator`, which handles all recurrence-related stuff. For `EventIterator`.

Specifically one issue here is recurring events and exceptions. To calculate all instances of a recurrence, you need to need to have all VEVENT objects that have a matching UID. UID's are not unique and when multiple events share a UID it means they belong together. There's no guarantee that events with the same UID appear in sequence.

8633brown:
Ahh interesting. So when calling the event iterator within the isInTimeRange function on [this](https://github.com/sabre-io/vobject/blob/ebe974884ef45cd6b62472b5942f8839ae48b5c9/lib/Component/VEvent.php#L37) line, if I was to call this on a recurring event without all other occurrences with the same uid it could give me the wrong result? As isInTimeRange only passes a single VEvent to the event iterator?
If that's correct I think I understand where the double pass comes in now.
 
Would it be an option to not filter recurring events. And only filter events which dont have an RRule?
Or as you said do a double pass and just put all VEvents with an RRule and the same UID into an array to do the second filter?
I guess I'm wondering if @evert you're still on the fence or if there is a chance this would be merged at all?

evert:
I don't consider myself a maintainer anymore, so ultimately it's up to others. However, I would lean towards 'no' because I feel it's a bit of an edge case, and somewhat imperfect due to the 'recurrence exception' issue.

I can see a need for this at some point, but maybe there's other ways to solve it. The VCardSplitter class for example stemmed from a similar issue with large VCard exports. This was a bit more grounded in real issues, because addressbook exports can be really massive due to pictures being stored in vcards. Possible in iCalendar too, but more rare.

For what it's worth, CalDAV does have a similar problem and also has a popular 'filter by timerange' feature. This is solved in sabre/dav by first creating indexes on events and finding out the 'first time an event occurs' and the 'last time an event occurs', and after that unprecise filter is applied, events are filtered _again_ with a slower precise filter that expands every occurrence.

Issue #464, opened 2019-07-05
Add "Eastern Time" to timezone mappings


alecpl:
Here's sample of iCal data:
```
BEGIN:VCALENDAR
PRODID:-//Microsoft Corporation//Outlook 10.0 MIMEDIR//EN
VERSION:2.0
METHOD:REQUEST
BEGIN:VTIMEZONE
TZID:Eastern Time
BEGIN:STANDARD
DTSTART:20161101T020000
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=1SU;BYMONTH=11
TZOFFSETFROM:-0400
TZOFFSETTO:-0500
TZNAME:Standard Time
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:20160301T020000
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=2SU;BYMONTH=3
TZOFFSETFROM:-0500
TZOFFSETTO:-0400
TZNAME:Daylight Savings Time
END:DAYLIGHT
END:VTIMEZONE
BEGIN:VEVENT
DTSTART;TZID="Eastern Time":20180430T100000
DTEND;TZID="Eastern Time":20180430T110000
...
```
According to one of our users it came from WebEX system.

I see that `Eastern Time (US & Canada)` already exists in exchangezones.php, but it does not match. Please, add an `Eastern Time` alias and also consider doing this for other timezones with similar problem.

PR #476, opened 2019-11-13
Add some more Exchange timezone aliases (#464)

alecpl:


evert:
The list of aliases is based on official microsoft sources. Do you have a source for this?

alecpl:
The linked ticket contains a real example. So called "WebEX invitation" contains `PRODID:-//Microsoft Corporation//Outlook 10.0 MIMEDIR//EN`.

I don't have another source.

evert:
Then I would discourage merging this as-is. It's not really possible to tell if that file was generated by exchange, it's just as likely that something modified it. All the files added to these aliases are all generated based on sources from real vendors, so at the very least it's sloppy to add it to the aliases file for exchange.

PR #483, opened 2020-01-02
Always run phpstan in travis

JeroenVanOort:
I think it would be better if phpstan is always ran for every php version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: php
-sudo: required
 php:
   - 5.5
   - 5.6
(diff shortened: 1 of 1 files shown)


staabm:
Do you have a case in mind which requires running it on all versions?

JeroenVanOort:
Based on the PHP version, different versions of dependencies may be used, which may or may not be compatible with project code. Static analysis seems to be a low-effort/high-gain tool to prevent issues.

In practice, however, phpstan is not compatible with 5.x, so it won't really work for now. It would work when support for these versions is dropped, as you suggested for 4.3. Would it be possible to open a branch for that version? So work can be done without bothering work for 4.2.*.

staabm:
You could just activate phpstan in the builds in which it works with `env: RUN_PHPSTAN="TRUE"`

staabm:
Thx

PR #484, opened 2020-01-02
Fix php 5.5 in travis

JeroenVanOort:


staabm:
Thx for the PR.

I guess we will drop support for php 5.x in the 4.3.x releases. 
Dont waste too much time on this old/dead php versions :-)

JeroenVanOort:
I hoped this would be an easy fix, but it seems that not all tests are compatible with the version of PHPUnit used when running php5.5.

PR #486, opened 2020-01-05
Add PHONE-NUMBER value type (used for TEL in vCard 3.0)

hanzi:
In vCard 3.0, `TEL` value should be of type `PHONE-NUMBER` ([Section 2.4.3](https://tools.ietf.org/html/rfc2426#section-2.4.3) / [Section 3.3.1](https://tools.ietf.org/html/rfc2426#section-3.3.1)).

So the following _should_ be a valid vCard:

```
BEGIN:VCARD
VERSION:3.0
TEL;TYPE=HOME;VALUE=PHONE-NUMBER:+123456789
END:VCARD
```

vObject currently throws an Exception when trying to parse that.

This PR adds the value type, and also throws it out when converting to vCard 4 (which expects `URI` type).


staabm:
Thx guys,

PR #487, opened 2020-01-07
Fixed phpstan level 1 errors

JeroenVanOort:
This fixes all phpstan level 1 errors and enables checking for them in CI. I'm working on a PR to also fix level 2 errors.

### DIFF
--- a/lib/Property/ICalendar/DateTime.php
+++ b/lib/Property/ICalendar/DateTime.php
@@ -338,8 +338,8 @@ class DateTime extends Property
         $messages = parent::validate($options);
         $valueType = $this->getValueType();
         $values = $this->getParts();
-        try {
-            foreach ($values as $value) {
+        foreach ($values as $value) {
+            try {
                 switch ($valueType) {
                     case 'DATE':
                         DateTimeParser::parseDate($value);
@@ -348,13 +348,13 @@ class DateTime extends Property
                         DateTimeParser::parseDateTime($value);
                         break;
                 }
+            } catch (InvalidDataException $e) {
+                $messages[] = [
+                    'level' => 3,
+                    'message' => 'The supplied value ('.$value.') is not a correct '.$valueType,
+                    'node' => $this,
+                ];
             }
-        } catch (InvalidDataException $e) {
-            $messages[] = [
-                'level' => 3,
-                'message' => 'The supplied value ('.$value.') is not a correct '.$valueType,
-                'node' => $this,
-            ];
         }
 
         return $messages;
--- a/lib/Property/ICalendar/DateTime.php
+++ b/lib/Property/ICalendar/DateTime.php
@@ -354,6 +354,7 @@ class DateTime extends Property
                     'message' => 'The supplied value ('.$value.') is not a correct '.$valueType,
                     'node' => $this,
                 ];
+                break;
             }
         }
 
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -322,14 +322,17 @@ class RRuleIterator implements Iterator
             return;
         }
 
+        $recurrenceHours = [];
         if (!empty($this->byHour)) {
             $recurrenceHours = $this->getHours();
         }
 
+        $recurrenceDays = [];
         if (!empty($this->byDay)) {
             $recurrenceDays = $this->getDays();
         }
 
+        $recurrenceMonths = [];
         if (!empty($this->byMonth)) {
             $recurrenceMonths = $this->getMonths();
         }
@@ -372,10 +375,12 @@ class RRuleIterator implements Iterator
             return;
         }
 
+        $recurrenceHours = [];
         if ($this->byHour) {
             $recurrenceHours = $this->getHours();
         }
 
+        $recurrenceDays = [];
         if ($this->byDay) {
             $recurrenceDays = $this->getDays();
         }
@@ -436,6 +441,7 @@ class RRuleIterator implements Iterator
             return;
         }
 
+        $occurrence = -1;
         while (true) {
             $occurrences = $this->getMonthlyOccurrences();
 
@@ -605,6 +611,7 @@ class RRuleIterator implements Iterator
         // If we got a byDay or getMonthDay filter, we must first expand
         // further.
         if ($this->byDay || $this->byMonthDay) {
+            $occurrence = -1;
             while (true) {
                 $occurrences = $this->getMonthlyOccurrences();
 
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,2 +1,4 @@
 parameters:
-  level: 0
+  level: 1
+  universalObjectCratesClasses:
+    - \Sabre\VObject\Component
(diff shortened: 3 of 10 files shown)


tests/VObject/ITip/BrokerTester.php
```diff
@@ -45,9 +45,7 @@ public function process($input, $existingObject = null, $expected = false)
 
         $vcal = Reader::read($input);
 
-        foreach ($vcal->getComponents() as $mainComponent) {
-            break;
-        }
+        $mainComponent = $vcal->getComponents()[0];
```
staabm: are we sure `$vcal->getComponents()` is never empty at this point?

lib/Recur/RRuleIterator.php
```diff
@@ -608,6 +614,7 @@ protected function nextYearly()
         // If we got a byDay or getMonthDay filter, we must first expand
         // further.
         if ($this->byDay || $this->byMonthDay) {
+            $occurrence = -1;
```
staabm: which error is fixed here?

lib/Property/ICalendar/DateTime.php
```diff
@@ -343,8 +343,8 @@ public function validate($options = 0)
         $messages = parent::validate($options);
         $valueType = $this->getValueType();
         $values = $this->getParts();
-        try {
-            foreach ($values as $value) {
+        foreach ($values as $value) {
```
staabm: which error is fixed here? it looks to me like the logic did change (before we stopped iteration on the first exception.. now iterate all values, no matter a exception is thrown in between)

tests/VObject/ITip/BrokerTester.php
JeroenVanOort: No, we are not. But neither were we before. This only makes sure that the variable is set. If `getComponents` is empty, an error would be thrown.

lib/Property/ICalendar/DateTime.php
JeroenVanOort: `$value` is used in the catch, which wouldn't be set if `$values` were empty.

This would indeed change the logic a bit; the user would receive all error messages instead of only the first one. I see that as a minor improvement.

Alternatively, `$value` could be set to null before the foreach, as that would also fix the phpstan error.

tests/VObject/ITip/BrokerTester.php
staabm: I guess with this change you traded one php warnings with another :-).
I think phpstan will yell at us for this new line on a higher level.

before your change we had a "maybe undefined variable".
with this change we might get a "undefined offset on array" warning

lib/Property/ICalendar/DateTime.php
staabm: feel free to suggest logic changes as separate PRs and concentrate on fixing just phpstan issues please.

otherwise we might introduce bugs

tests/VObject/ITip/BrokerTester.php
JeroenVanOort: That's correct.

I struggled with this, as the alternative would be to set `$mainComponent` to null before the loop. But then, we'd potentially call `->name` on null, which we can't do either.

There are many potential fixes for this problem: add an assert, add an if, and then some. But as this is a test, I'd like to keep focus on readability and not change too much. If phpstan will yell at us later, we'll fix it later. We might not even make it to the level at which this becomes a problem.

lib/Property/ICalendar/DateTime.php
JeroenVanOort: Ok. I will revert this and set `$value` to null before the foreach.

lib/Recur/RRuleIterator.php
JeroenVanOort: `$occurrence` may not be defined by the time it is used on line 654.

tests/VObject/ITip/BrokerTester.php
staabm: ohh I didn't realize we are in a test here. leave it as is then. thx for the note.

lib/Property/ICalendar/DateTime.php
```diff
@@ -343,6 +343,7 @@ public function validate($options = 0)
         $messages = parent::validate($options);
         $valueType = $this->getValueType();
         $values = $this->getParts();
+        $value = null;
```
DeepDiver1975: it is highly unusual to initialize the variable which is used within a foreach. From my POV it would be better to move the try/catch into the foreach and break out of the loop in case of exception

tests/VObject/ComponentTest.php
```diff
@@ -19,6 +19,7 @@ public function testIterate()
         $comp->add($sub);
 
         $count = 0;
+        $key = null;
```
DeepDiver1975: same as above - initializing the loop variable is highly uncommon

lib/Property/ICalendar/DateTime.php
JeroenVanOort: Ok, will fix this.

phpstan.neon
```diff
@@ -1,3 +1,5 @@
 parameters:
-  level: 0
+  level: 1
   bootstrap: %currentWorkingDirectory%/vendor/autoload.php
+  universalObjectCratesClasses:
```
staabm: What is this setting about?

JeroenVanOort: It declares to phpstan that any object of those classes is a universal crate, on which you can get and set any property even though it doesn't have those properties natively (through `__get` and `__set`). 'stdClass' is the most common example, but `Component` is one too.

If you wouldn't have this, you'd have to add an `@property` annotation for every possible value, like VEVENT, DTSTART, SUMMARY, etc.

staabm: Ahh I tried in the past defining all properties :-)

JeroenVanOort: I tried that too on my first attempt ;-)

JeroenVanOort:
May I ask why this is yet to be merged? As far as I know, all requested changes have been done.

staabm:
Please give us a few more days for final reviews.

@DeepDiver1975 any feedback left?

phil-davis:
@DeepDiver1975 this has been rebased just now on top of current master.
Please review again.

lib/Property/ICalendar/DateTime.php
phil-davis: @JeroenVanOort I don't see where you have reverted this logic change and done the `$value = null` or `$value = ''` thing?
What is the status?

JeroenVanOort: Soon after I made that comment, we discussed that setting a loop variable outside the loop should be avoided. That's why I kept the change, but added the break statement in order to prevent logic changes.

phil-davis: Thanks, I didn't notice the bonus `break;` at line 357!

phil-davis:
@staabm merge, or wait for @DeepDiver1975 to review again?

staabm:
Feel free to merge

phil-davis:
@JeroenVanOort thanks!
Feel free to improve the phpstan level even more ;)

PR #492, opened 2020-01-26
Apply code style

phil-davis:
- apply php-cs-fixer code style changes
- run php-cs-fixer in CI

This PR is on top of #491 

`php-cs-fixer` cleaned up a lot of PHPdoc where the parameter already had a type specified in the actual function call. I guess that is the "modern" way - specify the type in the function call and do not duplicate it in the PHPdoc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_script:
   - composer install
 
 script:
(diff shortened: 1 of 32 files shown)


phil-davis:
@staabm and/or @evert are you OK with making this much changes to keep php-cs-fixer happy?

PR #495, opened 2020-01-30
Refactored fqcn strings to ::class to allow checking with phpstan

JeroenVanOort:
I refactored the class strings to ::class references, to enable phpstan to check their existence. This makes it easier to prevent mistakes.

### DIFF
--- a/lib/Component/VCalendar.php
+++ b/lib/Component/VCalendar.php
@@ -37,15 +37,15 @@ class VCalendar extends VObject\Document
      * @var array
      */
     public static $componentMap = [
-        'VCALENDAR' => 'Sabre\\VObject\\Component\\VCalendar',
-        'VALARM' => 'Sabre\\VObject\\Component\\VAlarm',
-        'VEVENT' => 'Sabre\\VObject\\Component\\VEvent',
-        'VFREEBUSY' => 'Sabre\\VObject\\Component\\VFreeBusy',
-        'VAVAILABILITY' => 'Sabre\\VObject\\Component\\VAvailability',
-        'AVAILABLE' => 'Sabre\\VObject\\Component\\Available',
-        'VJOURNAL' => 'Sabre\\VObject\\Component\\VJournal',
-        'VTIMEZONE' => 'Sabre\\VObject\\Component\\VTimeZone',
-        'VTODO' => 'Sabre\\VObject\\Component\\VTodo',
+        'VCALENDAR' => self::class,
+        'VALARM' => VAlarm::class,
+        'VEVENT' => VEvent::class,
+        'VFREEBUSY' => VFreeBusy::class,
+        'VAVAILABILITY' => VAvailability::class,
+        'AVAILABLE' => Available::class,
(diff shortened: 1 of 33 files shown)


lib/Component/VCalendar.php
```diff
@@ -37,15 +37,15 @@ class VCalendar extends VObject\Document
      * @var array
```
staabm: Could we use a typehint like array<class-string> or similar now?

staabm:
```suggestion
     * @var array<string, class-string>
```

lib/Component/VCalendar.php
```diff
@@ -54,21 +54,21 @@ class VCalendar extends VObject\Document
...
```
staabm:
```suggestion
     * @var array<string, class-string>
```

lib/Component/VCard.php
```diff
@@ -39,7 +39,7 @@ class VCard extends VObject\Document
      * @var array
```
staabm:
```suggestion
     * @var array<string, class-string>
```

lib/Component/VCard.php
```diff
@@ -48,23 +48,23 @@ class VCard extends VObject\Document
...
```
staabm:
```suggestion
     * @var array<string, class-string>
```

tests/VObject/Component/VAvailabilityTest.php
```diff
@@ -236,7 +236,7 @@ public function testAvailableSubComponent()
 VCAL;
         $document = Reader::read($vcal);
 
-        $this->assertInstanceOf(__NAMESPACE__, $document->VAVAILABILITY->AVAILABLE);
+        $this->assertInstanceOf(Available::class, $document->VAVAILABILITY->AVAILABLE);
```
staabm: This looks like a error

lib/Component/VCalendar.php
JeroenVanOort: Yes, you can use them (as you suggested: `array<string, class-string>`), but they won't be enforced until phpstan level 3. Therefore, I doubt it is useful to add them now.

I have done some preliminary work to increase the phpstan level further, and when doing that we'll soon come across phpstan's demand to specify array's like you suggested.

tests/VObject/Component/VAvailabilityTest.php
JeroenVanOort: I did this on purpose, although it wasn't necessarily wrong before.

Before, this asserted that the object was an instance of `Component` as that is the namespace which the test is in. However, there is also a class Component which happens to share it's class path with the Component namespace. If I'd want to reference the Component class instead of the namespace, I'd need to add a `use` statement. Not too much trouble, but the test would most probably like to assert that the object is an instance of `Available`, which is what you'd expect. Therefore, I chose to do that instead.

lib/Component/VCalendar.php
staabm: I see. Thx for the clarification

PR #496, opened 2020-01-31
Various infra changes

JeroenVanOort:
These are some things I came across while working on this project. Please let me know if you have any questions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,13 +5,5 @@ tests/cov/
 tests/temp
 tests/.phpunit.result.cache
 
-#vim
-.*.swp
-
 # Development stuff
-testdata/
 .php_cs.cache
-.idea
-
-# OS X
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -8,13 +8,6 @@ tests/.phpunit.result.cache
 #vim
 .*.swp
 
-#binaries
-bin/phpunit
-bin/php-cs-fixer
-bin/phpstan
-bin/phpstan.phar
-bin/hoa
-
 # Development stuff
 testdata/
 .php_cs.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,6 @@ php:
 matrix:
   fast_finish: true
 
-install:
-  - composer require --dev phpstan/phpstan:^0.12
-
 before_script:
   - composer install
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_script:
   - composer install
 
 script:
-  - php vendor/bin/php-cs-fixer fix lib/ --dry-run --diff
+  - php vendor/bin/php-cs-fixer fix --dry-run --diff
   - php vendor/bin/phpstan analyse -c phpstan.neon lib tests
   - php vendor/bin/phpunit --configuration tests/phpunit.xml --coverage-clover=coverage.xml
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,10 @@ before_script:
   - composer install
 
 script:
-  - ./bin/php-cs-fixer fix lib/ --dry-run --diff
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -5,20 +5,5 @@ tests/cov/
...
-bin/phpunit
-bin/php-cs-fixer
-bin/phpstan
-bin/phpstan.phar
-bin/hoa
-
 # Development stuff
-testdata/
 .php_cs.cache
-.idea
```
phil-davis: This is handy to have.
When using PHPstorm IDE, it creates a folder called `.idea` 
Without this in `.gitignore` it is easy to accidentally commit the folder.

composer.json
```diff
@@ -38,7 +38,8 @@
     },
     "require-dev" : {
         "friendsofphp/php-cs-fixer": "~2.16.1",
-        "phpunit/phpunit" : "^7 || ^8"
+        "phpunit/phpunit" : "^7 || ^8",
+        "phpstan/phpstan": "^0.12"
```
phil-davis: @staabm I know that you wanted me to keep `phpstan` out of `composer.json`
But maybe that was just while we had PHP 7.0 support, and `phpstan` had to have a different version depending on the PHP version ...

.gitignore
JeroenVanOort: I think that files and folders created by the developer's environment should be in the [global .gitignore](https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore) of that developer. Otherwise, one would need to add so much more files and folders to support the many editors and other tools used.

phil-davis: How have I gone this long without anyone telling me this? And seen so many repos that have patterns like gthis listed in their `.gitignore`?

Thanks for educating me.

composer.json
staabm: as long as phpstan does not bump phpstan min version this should be fine

.travis.yml
```diff
@@ -8,17 +8,14 @@ php:
...
-
 before_script:
   - composer install
 
 script:
-  - ./bin/php-cs-fixer fix lib/ --dry-run --diff
-  - php ./bin/phpstan.phar analyse -c phpstan.neon lib tests
-  - ./bin/phpunit --configuration tests/phpunit.xml --coverage-clover=coverage.xml
-  
+  - php vendor/bin/php-cs-fixer fix -v --dry-run --stop-on-violation
```
staabm: We shouldnt stop on violations... having a full report is more helpfull IMO

composer.json
```diff
@@ -86,7 +87,20 @@
             "dev-master" : "4.0.x-dev"
         }
     },
-    "config" : {
-        "bin-dir" : "bin"
+    "scripts": {
```
staabm: Really like the scripts! Handy.

composer.json
```diff
@@ -86,7 +87,20 @@
...
+        "phpstan": [
+            "@php vendor/bin/phpstan analyse lib tests"
+        ],
+        "php-cs-fixer": [
```
staabm: I wouldnt name the fixer so we can change the tool beeing used without the need to chnage the devs workflow commands 
```suggestion
        "csfix": [
```

.travis.yml
JeroenVanOort: I thought this had to be there to make the travis build fail when violations are found. Will fix.

composer.json
JeroenVanOort: Will do.

phil-davis:
We should make similar changes in the other sabre-io repos to standardize them.

staabm:
> We should make similar changes in the other sabre-io repos to standardize them.

Agree, would be cool. Additionally I would love to have GithubActions instead of travis builds, because travis is sooo slow. 

In GithubActions we could also use https://github.com/staabm/annotate-pull-request-from-checkstyle then

PR #498, opened 2020-02-02
PHPstan level 2

JeroenVanOort:
There was no other way to make this happen than to make a custom PHPStan extension.

### DIFF
--- a/lib/Cli.php
+++ b/lib/Cli.php
@@ -317,13 +317,10 @@ HELP
     {
         $returnCode = 0;
 
-        switch (get_class($vObj)) {
-            case Component\VCalendar::class:
-                $this->log('iCalendar: '.(string) $vObj->VERSION);
-                break;
-            case Component\VCard::class:
-                $this->log('vCard: '.(string) $vObj->VERSION);
-                break;
+        if ($vObj instanceof Component\VCalendar) {
+            $this->log('iCalendar: '.(string) $vObj->VERSION);
+        } elseif ($vObj instanceof Component\VCard) {
+            $this->log('vCard: '.(string) $vObj->VERSION);
         }
 
         $warnings = $vObj->validate();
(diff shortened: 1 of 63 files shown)


lib/Cli.php
```diff
@@ -314,15 +313,15 @@ protected function showHelp()
...
      * @return int
      */
-    protected function validate(Component $vObj)
+    protected function validate(Document $vObj)
     {
         $returnCode = 0;
 
-        switch ($vObj->name) {
-            case 'VCALENDAR':
+        switch (get_class($vObj)) {
```
staabm: I think this is a BC break, when people register custom classes in the component map?

JeroenVanOort: It would be, yes. But how is registering custom classes in the component map supported?

staabm: Hmm just by writing into the public propertyMap

JeroenVanOort: What I can do is refactor this to use `instanceof`. Would that be a satisfactory solution?

staabm: I think this can/will only work by name and the corresponding keys... no way to mal this over the fqcn

staabm: After you reverted the changes related to the Mapping logic, we could land the rest of the PR.

JeroenVanOort: When someone registers a custom class in the component map, wouldn't (or shouldn't) that always be an extension of the original class? If so, things would work just fine if we were to use `instanceof`.

I believe there are already many places where references to classes are used as opposed to their names, so I would be inclined to say that registering custom classes that don't extend the original class won't work anyway.

staabm: @DeepDiver1975 @evert could we get your opinion on this one? I have no experience in this area..

evert: It feels like a low-risk change, especially since this is the cli tool

staabm: thx @evert . do you feel the same for the similar change in https://github.com/sabre-io/vobject/pull/498/files#diff-fb8f68f07cb5075b44bf17c664d7c9bdR37 ?

evert: Harder to say! Might be a BC break, as I don't have 100% visibility on how people use and extend this. In theory it's possible to create other components that are a VTODO but not extend the base class. I would somewhat expect that other bugs _might_ crop up for these people Overall it seems like a positive change, but not 100% sure if its gonna be an issue for folks

phil-davis:
@JeroenVanOort I merged PR #500 that "standardizes" more CI stuff.
Sorry, now you have conflicts.
Please rebase and resolve...

lib/Cli.php
JeroenVanOort: I've refactored the comparisons with `get_class` to `instanceof` as to allow custom extensions.

phil-davis:
@JeroenVanOort this has been sitting for a while and now has conflicts. Are you able to rebase and resolve them again?

PR #508, opened 2020-06-23
Require sabre/xml 2.2

phil-davis:
`sabre/xml` 2.2 is the release that supports PHP 7.4.

So maybe we should require that in `composer.json` to make sure that at least 2.2 comes automagically with the current `vobject`

Probably it will get fetched anyway, just being sure.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require" : {
         "php"          : "^7.1",
         "ext-mbstring" : "*",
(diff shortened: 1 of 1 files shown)


staabm:
composer.json should define the minimum required version. Composer will itself try to fetch the latest possible.

We define the min version though, because then we get a bigger range of possible versions when combined with other libs.

phil-davis:
Closing - not needed.

PR #514, opened 2020-10-03
Release 4.3.2

phil-davis:
Prepare release 4.3.2 so that vobject will advertise PHP 8.0 support.

This will help projects that use this library to try PHP 8.0


staabm (approved):
Go phil go phil  <3

PR #522, opened 2020-11-09
Release 4.3.3

phil-davis:
Note: this release removes an ancient (1992) and "never used" timezone that has been deprecated upstream for ages and has now been removed upstream in the "standard" timezone database, and in PHP's list of known timezones. There really should be absolutely nobody using it. So I am making this just a patch release because it fixes this for people who are updating to the latest PHP `7.*.*` patch releases, and the "backward-compatibility break" is so minor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ChangeLog
 =========
 
+4.3.3 (2020-11-09)
+------------------
+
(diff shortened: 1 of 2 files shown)


phil-davis:
CI is running: https://travis-ci.org/github/sabre-io/vobject/builds/742349040
Travis has not updated the status here in GitHub.

PR #523, opened 2020-11-14
Fix setting properties with group assignment

mstilkerich:
Setting properties that are part of a group is broken since version 4.0.0. The `Component::remove()` method for the specific case of a property given by its name (as string) that includes a group prefix will throw an exception. However, I believe the exception is only intended for the usage of this function where the property to remove is given as an object reference, not as a string, where the given property/component is not a child of component the remove() method is called on.

Example code to trigger the exception:
```php
<?php

use Sabre\VObject;
use Sabre\VObject\Component\VCard;
include_once dirname(__FILE__) . "/vendor/autoload.php";

$vcard = new VObject\Component\VCard([
        'VERSION' => '3.0',
        'item2.X-ABLabel' => "Foo", // triggers exception
]);

$vcard->{"item1.X-ABLabel"} = "SpecialLabel"; // also triggers exception

print $vcard->serialize();
```

The exception will be triggered for both the usage in the constructor as well as by a separate set operation.

### DIFF
--- a/lib/Component.php
+++ b/lib/Component.php
@@ -160,9 +160,9 @@ class Component extends Node
                     return;
                 }
             }
-        }
 
-        throw new \InvalidArgumentException('The item you passed to remove() was not a child of this component');
+            throw new \InvalidArgumentException('The item you passed to remove() was not a child of this component');
+        }
     }
 
     /**
--- a/tests/VObject/ComponentTest.php
+++ b/tests/VObject/ComponentTest.php
@@ -72,6 +72,23 @@ class ComponentTest extends TestCase
         $this->assertEquals(null, $email3[0]->group);
     }
 
+    public function testAddGroupProperties()
+    {
+        $comp = new VCard([
+            'VERSION' => '3.0',
+            'item2.X-ABLabel' => "item2-Foo",
+        ]);
+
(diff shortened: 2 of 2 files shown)


DeepDiver1975 (changes requested):
Thanks a lot for your contribution!

Can I ask you to add a unit test for this case? :+1:

phil-davis (commented):
Code looks reasonable. Just needs unit test(s) so this does not get broken again in future.

mstilkerich:
Hello, I added a test case.

mstilkerich:
Hello, I added the requested test. Is there still a problem?

phil-davis:
With the new test case on `master`
```
There was 1 error:

1) Sabre\VObject\ComponentTest::testAddGroupProperties
InvalidArgumentException: The item you passed to remove() was not a child of this component

/home/phil/git/sabre-io/vobject/lib/Component.php:165
/home/phil/git/sabre-io/vobject/lib/Component.php:483
/home/phil/git/sabre-io/vobject/tests/VObject/ComponentTest.php:82

ERRORS!
Tests: 1721, Assertions: 2639, Errors: 1.
Script phpunit --configuration tests/phpunit.xml handling the phpunit event returned with error code 2

```

That's good, the test fails. And passes with the fix.

phil-davis (approved):

PR #525, opened 2020-11-27
Add .gitattributes

fezfez:
This will drop downloaded package size drastically

before : 219 items, size 1,3 MB
after : 106 items, size 580,2 kB


staabm (approved):
Thanks

PR #528, opened 2020-12-13
Use latest php-cs-fixer 2.17.1

phil-davis:
The latest php-cs-fixer finds some new things. Code changes were needed in https://github.com/sabre-io/dav/pull/1316 for `sabre/dav`. We might as well consistently make sure to use at least this version 2.17.1 in all repos so that we are doing consistent code-style checking.

It complains about:
- underscore `_` in method names. 
- use of `die()` (wants it to be `exit()`)
- blank lines in switch-case

1) adjust code to make php-cs-fixer happy
2) Update windowszones timezone data to 2020-12-13
3) Fix various typos

### DIFF
--- a/bin/fetch_windows_zones.php
+++ b/bin/fetch_windows_zones.php
@@ -1,13 +1,12 @@
 #!/usr/bin/env php
 <?php
 
-$windowsZonesUrl = 'http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml';
+$windowsZonesUrl = 'https://raw.githubusercontent.com/unicode-org/cldr/master/common/supplemental/windowsZones.xml';
 $outputFile = __DIR__.'/../lib/timezonedata/windowszones.php';
 
 echo 'Fetching timezone map from: '.$windowsZonesUrl, "\n";
 
 $data = file_get_contents($windowsZonesUrl);
-
 $xml = simplexml_load_string($data);
 
 $map = [];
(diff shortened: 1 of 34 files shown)


lib/timezonedata/windowszones.php
```diff
@@ -10,7 +10,7 @@
  * @license http://sabre.io/license/ Modified BSD License
  */
 
-return  [
+return [
```
phil-davis, line 13: This file says that it is auto-generated. But it has not been updated from the "source" any time recently.
php-cs-fixer did not like having 2 spaces here, so I made the edit anyway.
If someone re-generates the file from upstream some day, they might need to do a small edit like this to keep php-cs-fixer happy.

phil-davis: Oh, actually it is "auto-generated" from out own script. https://github.com/sabre-io/vobject/blob/master/bin/fetch_windows_zones.php

So it does not matter. I will update that script also...

phil-davis:
@DeepDiver1975 @staabm please review. I do not feel comfortable merging it directly - there are quite a few lines of diff.

bin/fetch_windows_zones.php
```diff
@@ -44,6 +43,6 @@
 
 echo "Formatting\n";
 
-exec(__DIR__.'/sabre-cs-fixer fix '.escapeshellarg($outputFile));
+exec(__DIR__.'/../vendor/bin/php-cs-fixer fix '.escapeshellarg($outputFile));
```
staabm, line 46: hmm not sure its correct that this path changed? could you verify this bin still works?

staabm (approved):
except the one mentioned line, it looks good to me.

thanks!

bin/fetch_windows_zones.php
phil-davis: I ran `fetch_windows_zones.php` locally from the root of the repo in order to update the windows zones file.
That is how I found that the old `sabre-cs-fixer` is no longer, and made it work by doing this.

PR #532, opened 2021-02-10
Make use of until parameter in nextMonthly function

jaircuevajunior:
Hello there,

I've felt the need of this PR becauase of a buggy event:
FREQ=MONTHLY;
INTERVAL=1;
BYMONTHDAY=2;
BYDAY=2WE;
WKST=WE;
UNTIL=20211214T000000Z

The "Until" was not being taken into account, so it was falling into the 10000 year problem.

### DIFF
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -466,13 +466,12 @@ class RRuleIterator implements Iterator
 
             // This goes to 0 because we need to start counting at the
             // beginning.
-	    $currentDayOfMonth = 0;
+            $currentDayOfMonth = 0;
 
-            // For some reason the "until" parameter was not being taken into
-	    // account, that's why the workaround of the 10000 year bug was
-	    // needed at all.
-            // Let's stop it before the "until" parameter date arrives.
-            if ($this->currentDate->getTimestamp() >= $this->until->getTimestamp()){
+            // For some reason the "until" parameter was not being used here,
+            // that's why the workaround of the 10000 year bug was needed at all
+            // let's stop it before the "until" parameter date
+            if ($this->currentDate->getTimestamp() >= $this->until->getTimestamp()) {
                 return;
             }
 
@@ -483,7 +482,7 @@ class RRuleIterator implements Iterator
 
                 return;
             }
-	}
+        }
 
         $this->currentDate = $this->currentDate->setDate(
             (int) $this->currentDate->format('Y'),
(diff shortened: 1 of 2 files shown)


staabm:
Thanks for the PR.

Could you add unit test which fails without your fix?

jaircuevajunior:
The problem is that this is a performance issue...
If you get this rule, you'll realize it's an impossible rule, because of the `BYMONTHDAY=2;`.
When the object receives an impossible rule, it will loop through all the possibilities until it reaches this breakpoint:
```php
            // To prevent running this forever (better: until we hit the max date of DateTimeImmutable) we simply
            // stop at 9999-12-31. Looks like the year 10000 problem is not solved in php ....
            if ($this->currentDate->getTimestamp() > 253402300799) {
                $this->currentDate = null;

                return;
            }
```

But it could stop the loop as soon as it arrives the `until` parameter as proposed in my PR:
```php
            // For some reason the "until" parameter was not being used here,
            // that's why the workaround of the 10000 year bug was needed at all
            // let's stop it before the "until" parameter date
            if ($this->until && $this->currentDate->getTimestamp() >= $this->until->getTimestamp()) {
                return;
            }
```

In both cases the result is gonna be the same, but the process will be much faster with my purposed fix.

NOTE: Obviously, this is only working when the `until` parameter is set, otherwise it'll run untill the 10000 year as normally expected.

Now i have runned the tests, found an error and fixed it, but I don't know the right way of creating a test for this case since it's a performance issue... I don't want to create a test using the time as a parameter to fail the test... any idea?

lib/Recur/RRuleIterator.php
```diff
@@ -468,6 +468,13 @@ protected function nextMonthly()
             // beginning.
             $currentDayOfMonth = 0;
 
+            // For some reason the "until" parameter was not being used here,
+            // that's why the workaround of the 10000 year bug was needed at all
+            // let's stop it before the "until" parameter date
+            if ($this->until && $this->currentDate->getTimestamp() >= $this->until->getTimestamp()) {
+                return;
```
phil-davis, line 475: Codecov reports that there is no test that gets inside this `if` statement,
So you could add a test case that has `until` set. That will verify that this `if` statement can evaluate `true` and return appropriately.

jaircuevajunior: Done.

phil-davis (approved):
LGTM

Issue #534, opened 2021-05-17
EventIterator returns wrong end endTime


floerke:
Hi,

the EventIterator returns a wrong endTime for
- a recurring event
- with an exception and a different endTime

**Example:**
_Everyday from 9am-10am, except on 20210519 from 9am-12am_

```
BEGIN:VCALENDAR
PRODID:-//Mozilla.org/NONSGML Mozilla Calendar V1.1//EN
VERSION:2.0
BEGIN:VTIMEZONE
TZID:Europe/Berlin
BEGIN:DAYLIGHT
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
TZNAME:CEST
DTSTART:19700329T020000
RRULE:FREQ=YEARLY;BYMONTH=3;BYDAY=-1SU
END:DAYLIGHT
BEGIN:STANDARD
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
TZNAME:CET
DTSTART:19701025T030000
RRULE:FREQ=YEARLY;BYMONTH=10;BYDAY=-1SU
END:STANDARD
END:VTIMEZONE

BEGIN:VEVENT
CREATED:20210517T073701Z
LAST-MODIFIED:20210517T095958Z
DTSTAMP:20210517T095958Z
UID:e47c8ee2-5545-4c15-8fe5-7dd1960a4446
SUMMARY:9Uhr
RRULE:FREQ=DAILY
DTSTART;TZID=Europe/Berlin:20210517T090000
DTEND;TZID=Europe/Berlin:20210517T100000
TRANSP:OPAQUE
X-MOZ-GENERATION:1
END:VEVENT

BEGIN:VEVENT
CREATED:20210517T073809Z
LAST-MODIFIED:20210517T073833Z
DTSTAMP:20210517T073833Z
UID:e47c8ee2-5545-4c15-8fe5-7dd1960a4446
SUMMARY:9Uhr
RECURRENCE-ID;TZID=Europe/Berlin:20210519T090000
DTSTART;TZID=Europe/Berlin:20210519T090000
DTEND;TZID=Europe/Berlin:20210519T120000
TRANSP:OPAQUE
SEQUENCE:1
END:VEVENT
END:VCALENDAR
```

Results in:
_Start 9am, Ends 10am on 20210519_

I have noticed this because my thunderbird/lightning showing wrong freebusy-times in this case using nextcloud. Your FreeBusyGenerator.php is using the EventIterator to calculate the FreeBusy hours.

[...]

PR #538, opened 2021-08-04
EventIterator returns wrong endTime (#534)

floerke:
Fix for #534


phil-davis (approved):
Squashed and rebased to get fresh CI - LGTM

Issue #540, opened 2021-09-10
Reordering of the attendees should not be a signitifcant change


floerke:
If a server (using sabre) receives a VEVENT update it will send mails to all attendees. Not for all modified events, but for events with a "significant change". The calculation of "signifcant change" is done by vobject. 

I have trouble with to many "Invitation: ..."-Mails for events that has not been changed. 

Investigating this I have seen:
- the events were modified by the client and sent to the server (eg. to save the acknowledgment of an alarm).
- the client has changed the order (not the attendees itself) of the attendees in the modified event
- vobject calculated a "significant change"
- the server sent the email

I think everything is correct, except, the reordering is not a significant change. Why the client desires to reorder the attendees? I don't know, but I think this should not be a significant change.

phil-davis:
Fixed by #541

PR #541, opened 2021-09-10
Reordering of the attendees should not be a signitifcant change (#540)

floerke:
https://github.com/sabre-io/vobject/issues/540

If a server (using sabre) receives a VEVENT update it will send mails to all attendees. Not for all modified events, but for events with a "significant change". The calculation of "signifcant change" is done by vobject.

I have trouble with to many "Invitation: ..."-Mails for events that has not been changed.

Investigating this I have seen:

- the events were modified by the client and sent to the server (eg. to save the acknowledgment of an alarm).
- the client has changed the order (not the attendees itself) of the attendees in the modified event
- vobject calculated a "significant change"
- the server sent the email

I think everything is correct, except, the reordering is not a significant change. Why the client desires to reorder the attendees? I don't know, but I think this should not be a significant change.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -547,9 +547,13 @@ class Broker
                 // properties changed in the event, or simply if there's a
                 // difference in instances that the attendee is invited to.
 
+                $oldAttendeeInstances = array_keys($attendee['oldInstances']);
+                $newAttendeeInstances = array_keys($attendee['newInstances']);
+
                 $message->significantChange =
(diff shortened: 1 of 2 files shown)


staabm:
Sounds reasonable. Could you add a unit test?

floerke:
Hi Markus, I added a new unittest...

phil-davis (approved):
Rebased and squashed to get fresh CI. LGTM.

Issue #542, opened 2021-09-10
Reordering of vevent in a recurring event with exceptions should not be a significant change


floerke:
Related to #540 

If a server (using sabre) receives a VEVENT update it will send mails to all attendees. Not for all modified events, but for events with a "significant change". The calculation of "signifcant change" is done by vobject.

I have trouble with to many "Invitation: ..."-Mails for events that has not been changed.

Investigating this I have seen:

-    the events were modified by the client and sent to the server (eg. to save the acknowledgment of an alarm).
-    the client has changed the order (not the vevents itself) of the venets in the modified event
-    vobject calculated a "significant change" (by SignificantChangeHash)
-    the server sent the email

I think everything is correct, except, the reordering is not a significant change. Why the client desires to reorder the vevents? I don't know, but I think this should not be a significant change.

phil-davis:
Fixed by #543

PR #543, opened 2021-09-10
Reordering of vevent should not be a significant change (#542)

floerke:
Related to #540

If a server (using sabre) receives a VEVENT update it will send mails to all attendees. Not for all modified events, but for events with a "significant change". The calculation of "signifcant change" is done by vobject.

I have trouble with to many "Invitation: ..."-Mails for events that has not been changed.

Investigating this I have seen:

- the events were modified by the client and sent to the server (eg. to save the acknowledgment of an alarm).
-  the client has changed the order (not the vevents itself) of the venets in the modified event
-  vobject calculated a "significant change" (by SignificantChangeHash)
-  the server sent the email

I think everything is correct, except, the reordering is not a significant change. Why the client desires to reorder the vevents? I don't know, but I think this should not be a significant change.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -820,7 +820,10 @@ class Broker
         $instances = [];
         $exdate = [];
 
+        $significantChangeEventProperties = [];
+
         foreach ($calendar->VEVENT as $vevent) {
+            $eventSignificantChangeHash = '';
             $rrule = [];
 
             if (is_null($uid)) {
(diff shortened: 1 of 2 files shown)


phil-davis:
I will rebase and resolve the conflict...

phil-davis (approved):
Rebased, squashed, conflicts resolved. I fixed a typo of "significant" in a previous test case. And added a comment in the new test case (it took me a bit of staring to see what was the actual difference in the 2 VEVENT declarations)

LGTM

PR #548, opened 2021-10-29
Allow easier extension of the timezone guessing

heiglandreas:
This will ease customization of timezone-guessing as it is now gets easier to extend that process with own implementations (as long as they implement the appropriate interface)

This is espechially necessary when wanting to actually guess a timezone via the rules defined in the VTIMEZONE-entry (which is currently not done)

This change does not introduce any new logic. It just rearranges the already existing code in a different way to allow creating own TimezoneGuessers that can extend the functionality like this:

```php
class BuesingenGuesser implements TimezoneGuesser
{
    public function guess(VTimeZone $vtimezone, bool $throwWhenUnsure = false): ?DateTimeZone
    {
        return new DateTimeZone('Europe/Busingen');
    }
}
TimezoneUtil::addTimeZoneGuesser(new BuesingenGuesser());
```

Now every time a timezone is not resolvable the timezone will be resolved to `Europe/Busingen` instead of `UTC` (or whatever else is set via `date_default_timezone_set()`)

### DIFF
--- a/lib/TimeZoneUtil.php
+++ b/lib/TimeZoneUtil.php
@@ -2,6 +2,16 @@
 
 namespace Sabre\VObject;
 
+use DateTimeZone;
+use InvalidArgumentException;
+use Sabre\VObject\TimezoneGuesser\FindFromOffset;
+use Sabre\VObject\TimezoneGuesser\FindFromTimezoneIdentifier;
+use Sabre\VObject\TimezoneGuesser\FindFromTimezoneMap;
+use Sabre\VObject\TimezoneGuesser\GuessFromLicEntry;
+use Sabre\VObject\TimezoneGuesser\GuessFromMsTzId;
+use Sabre\VObject\TimezoneGuesser\TimezoneFinder;
+use Sabre\VObject\TimezoneGuesser\TimezoneGuesser;
+
 /**
  * Time zone name translation.
  *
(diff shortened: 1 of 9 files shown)


lib/TimeZoneUtil.php
```diff
@@ -263,3 +168,6 @@ public static function getIdentifiersBC()
         return include __DIR__.'/timezonedata/php-bc.php';
     }
 }
+
+TimeZoneUtil::addTimezoneGuesser(new GuessFromLicEntry());
```
staabm, line 172: Could we move this into a constructor?

staabm (commented):
Really like this direction. Do we have/need test coverage for this code?

lib/TimeZoneUtil.php
heiglandreas: As there *is* no constructor due to the static nature that is a bit of an issue. I'd love to have that in a constructor but that would mean that the whole class and it's signature needs to change. Any ideas how to solve that?

staabm: Maybe alternatively move it into a procedural-only file and include it via composer-autoload-file?

heiglandreas: I've just reworked the whole class into using an instance and exposing the relevant functions via some static functions and a singleton-pattern.

I additionally also reworked the timezone finding into own classes to make extending that possible as well.

heiglandreas:
Should be good by now..

lib/TimeZoneUtil.php
```diff
@@ -14,94 +24,39 @@
...
 {
-    public static $map = null;
+    private static $instance = null;
 
-    /**
-     * List of microsoft exchange timezone ids.
-     *
-     * Source: http://msdn.microsoft.com/en-us/library/aa563018(loband).aspx
-     */
-    public static $microsoftExchangeMap = [
```
staabm, line 24: I am not sure we can drop this field, as other code might rely on it beeing it.

there is more public api changed with this PR in a non-BC manner

lib/TimeZoneUtil.php
```diff
@@ -14,94 +24,39 @@
  */
 class TimeZoneUtil
 {
-    public static $map = null;
+    private static $instance = null;
```
staabm, line 27:
```suggestion
    /**
     * @var self
     */
    private static $instance = null;
```

lib/TimeZoneUtil.php
```diff
@@ -14,94 +24,39 @@
...
-        36 => 'America/Edmonton',
-        38 => 'America/Phoenix',
-        12 => 'America/Denver', // Best guess
-        13 => 'America/Los_Angeles', // Best guess
-        14 => 'America/Anchorage',
-        15 => 'Pacific/Honolulu',
-        16 => 'Pacific/Midway',
-        39 => 'Pacific/Kwajalein',
-    ];
+    private $timezoneFinders = [];
```
staabm, line 31:
```suggestion
    /**
     * @var TimezoneFinder[]
     */
    private $timezoneFinders = [];
```

staabm (commented):
just leaving a few comments. but lets wait what the other devs think.

lib/TimeZoneUtil.php
heiglandreas: Right!. I'll readd those other public items to stay backwards compatible.

Main question though is whether I shall do it in a way that allows them to be easily removed at a later stage (meaning to duplicate some code) or to have as little duplicate code as possible.

Personally I'd opt for the former version but am happy to modify it to the later one.

phil-davis:
@heiglandreas the code style is failing. It is easily fixed by:

`composer cs-fixer`

and then commit and push the changes it makes.

phil-davis:
https://github.com/sabre-io/vobject/runs/4101303777?check_suite_focus=true
`Class Sabre\VObject\TimeZoneUtil referenced with incorrect case`

The code analysis is fussy. Please fix that up.

Note: I released https://github.com/sabre-io/vobject/releases/tag/4.3.6 an hour ago. But when we get this PR sorted out, it will be easy to make a 4.3.7 release.

tests/VObject/TimeZoneUtilTest.php
```diff
@@ -8,8 +8,8 @@ class TimeZoneUtilTest extends TestCase
 {
     public function setUp(): void
     {
-        // clearing the tz cache
-        TimeZoneUtil::$map = null;
+        // clearning the tz cache
```
phil-davis, line 11:
```suggestion
        // cleaning the tz cache
```

Do we even need this comment? It says no more than what the code is "obviously" (to me) doing.

lib/TimeZoneUtil.php
```diff
@@ -14,17 +24,136 @@
...
+                    }
+                }
+            }
+        }
+
+        if ($failIfUncertain) {
+            throw new InvalidArgumentException('We were unable to determine the correct PHP timezone for tzid: '.$tzid);
+        }
+
+        // If we got all the way here, we default to UTC.
```
phil-davis, line 109:
```suggestion
        // If we got all the way here, we default to whatever has been set as the PHP  default timezone.
```

phil-davis (commented):
Just a couple of tiny comments.

phil-davis:
@staabm @evert @DeepDiver1975 please review and comment.
This would be at least a minor version bump (new feature - way of processing timezones). Please think if there is any remaining BC break that would require a major version.

staabm (commented):
lgtm

staabm (approved):

phil-davis:
I squashed and rebased to get up-to-date clean CI.

phil-davis (approved):
LGTM. There is some discussion in PR #553 about tests to cover the functions `addTimezoneGuesser` and `addTimezoneFinder` but that can be left until later.

PR #549, opened 2021-11-03
sync ci.yml to match other repos

phil-davis:
Just a few name changes so that this matches with https://github.com/sabre-io/http and we can copy-paste to each repo without constantly wondering why the name text is slightly different everywhere.


staabm:
fyi, for a few weeks workflow can be re-used across repos

https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

PR #553, opened 2021-11-15
Add timezone function tests

phil-davis:
This adds test cases for `addTimezoneGuesser` and `addTimezoneFinder` on top of #548 

I just tests that the functions can be called without throwing an exception!  That at least verifies that the functions still exist and can be called with the tested parameters.

There is not `getTimezoneGuessers()` function, so it is not immediately easy to know if the add function did anything. To fully test we would have to provide some new-and-different guesser(s), add it, then do some calls to `findTimeZone` and observe a difference in behavior. I'm not sure how much of that we should try and do.

But I am also not sure if these "mindless" test scenarios are worth having.

### DIFF
--- a/lib/TimeZoneUtil.php
+++ b/lib/TimeZoneUtil.php
@@ -2,6 +2,16 @@
 
 namespace Sabre\VObject;
 
+use DateTimeZone;
+use InvalidArgumentException;
+use Sabre\VObject\TimezoneGuesser\FindFromOffset;
+use Sabre\VObject\TimezoneGuesser\FindFromTimezoneIdentifier;
+use Sabre\VObject\TimezoneGuesser\FindFromTimezoneMap;
+use Sabre\VObject\TimezoneGuesser\GuessFromLicEntry;
+use Sabre\VObject\TimezoneGuesser\GuessFromMsTzId;
+use Sabre\VObject\TimezoneGuesser\TimezoneFinder;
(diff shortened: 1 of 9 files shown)


phil-davis:
@staabm @heiglandreas
should I add the last 2 commits here to PR #548 ?
It covers a few more little things that codecov was complaining about.
But maybe the  "tests" that I wrote for the add timezone guesser functions are rather mindless/useless?

Opinion please!

staabm:
hmm `addFinder` and `addGuesser` are simple adder/setter methods which cannot throw.
I don't see where/when a exception could/would be thrown?

phil-davis:
> hmm `addFinder` and `addGuesser` are simple adder/setter methods which cannot throw. I don't see where/when a exception could/would be thrown?

That's the challenge of these simple functions! If someone changes it some day and breaks it (in a way that throws and exception, or they rename or delete the function) then the test will fail. So it protects against obvious "regressions" like that.

I'm not sure how to really do better than that, without creating some new example guesser and using it...

staabm:
I would tackle this problem not on a per method basis, but with a BC check which covers the whole codebase/project regarding BC, e.g. https://github.com/Roave/BackwardCompatibilityCheck

phil-davis:
I added `testEmptyTimeZone` to PR #548 which is now merged.

The other tests are not that useful.

PR #560, opened 2022-01-03
Type property parameters via phpdoc

ChristophWurst:
And update the property constructor to better reflect that the
$parameters are an array of values, not instances of the parameter
class.

I suppose this can be seen as part of https://github.com/sabre-io/dav/issues/1368.

Context: I converted some Nextcloud code into strict code and the section with properties/parameters was not understood by PHPStorm until I made these changes.

### DIFF
--- a/lib/Property.php
+++ b/lib/Property.php
@@ -37,7 +37,7 @@ abstract class Property extends Node
     /**
      * List of parameters.
      *
-     * @var array
+     * @var Parameter[]
      */
     public $parameters = [];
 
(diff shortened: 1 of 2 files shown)


lib/Property.php
```diff
@@ -37,7 +37,7 @@ abstract class Property extends Node
     /**
      * List of parameters.
      *
-     * @var array
+     * @var Parameter[]
```
staabm, line 40: I think we should go one step further:
```suggestion
     * @var array<string, Parameter>
```

ChristophWurst: I like that as well, yet I'm not sure how "official" that phpdoc syntax is.

staabm: its kind of proprietary syntax which is used in psalm, phpstan and most static analysis tooling.

we could also use a more tool targetting syntax, so we don't kill the "offical form":

```suggestion
     * @phpstan-var array<string, Parameter>
     * @var Parameter[]
```

ChristophWurst: If you prefer this then I can add it

PR #561, opened 2022-01-14
PHP 8.1 deprecated support for null values in its APIs

mstilkerich:
When doing an addressbook query report (using Baikal, but I guess it won't matter) on PHP 8.1, PHP raises deprecation notices because strtoupper is called with a null value. This is fixed in this PR.


phil-davis (approved):
Looks reasonable.

@staabm what should we do about unit tests for this sort of stuff?
Do we try to engineer a unit test scenario where `$child->group` is `null` and demonstrate that it "fails" on PHP 8.1 with a deprecation notice?

lib/Component.php
```diff
@@ -238,7 +238,7 @@ public function select($name)
                 return array_filter(
                     $result,
                     function ($child) use ($group) {
-                        return $child instanceof Property && strtoupper($child->group) === $group;
+                        return $child instanceof Property && strtoupper((string) $child->group) === $group;
```
staabm, line 241: I would prefer an explicit null check instead of a cast

staabm:
> Do we try to engineer a unit test scenario where `$child->group` is `null` and demonstrate that it "fails" on PHP 8.1 with a deprecation notice?

IMO this should be covered by static analysis. so we e.g. might increase the phpstan level until it reports this kind of errors

lib/Component.php
staabm: it would also be usefull if we could come up with a more fine grained type then `array` for the `children` property

mstilkerich: I actually used the case because this is also done in other places in the existing code and IMO it's the best readable option when still needing to support PHP 5.5 as you intend to by the README. Anyway, I changed the code to match your preference.
I also adapted the array type, also I don't see how it helps here. What could help is declaring the `$group` property as nullable in the Property class, but phpstan does not seem to care, maybe because it is set to a very basic level.

mstilkerich:
If you enabled reporting of E_DEPRECATE errors when running the unit tests, these error would have shown up. I did so and fixed the other reported passing of null to PHP APIs where a string is expected.

phil-davis:
There are couple of minor code-style things the cs-fixer is complaining about.

tests/phpunit.xml.bak
```diff
@@ -0,0 +1,24 @@
+<phpunit
```
phil-davis, line 1: this looks like is was accidentally committed

mstilkerich: right, removed this file again.

mstilkerich:
Hello, I believe I fixed all of your findings. If there’s still something open you are waiting for please let me know.

mstilkerich:
Hello, just in case anyone cares, it has been kind of frustrating experience for me trying to contribute to the sabre projects. At this point, I have three open PRs which have stayed without feedback for a long time, in part for years, or like this one where I was first asked to fix style violations and stuff just to be ignored after having done all that. Now that my CI system has also moved to PHP 8.1, I need to use a patched version of Baikal because my tests fail as a result of the issue fixed in this PR. As you claim to support PHP8 and it is becoming more and more widespread, I have no clue why you show no interested in fixing these simple issues.

Anyway, I have to say I am pretty discouraged from submitting any more contributions to the sabre projects after these experiences. Which is a shame given that sabre is the leading open source DAV server implementation.

staabm:
I can totally see that the process is frustrating. 

We have a similar problem a lot other OSS projects have. We don't have full time maintainers and everyone involved is doing the best he/she can in freetime. 

If your business depends on the project you should consider supporting the people working on it, so we can afford spending more time on it.

mstilkerich:
My business doesn’t. My OSS project (a carddav client) does in that I attempt to verify its interoperability with sabre dav.

Contributing fixes to problems I discover is the way I can support you, at least that’s what I thought.

phil-davis:
@mstilkerich I  am working on https://github.com/sabre-io/xml/issues/215 right now. I just released a `sabre/uri` that has all PHP types declared (there was not much to do there). I am about to release a major version of `sabre/xml` that requires a minimum of PHP 7.4 and has type declarations everywhere.

Once I think that is working OK, I will move on to the other repos, starting with the ones that do not depend on too much other stuff!

And I should be able to get your PR sorted out and released as a patch version, before doing bigger changes here in vobject. You code should run OK across all the currently-supported PHP  versions 7.1 through 8.1

phil-davis:
@mstilkerich can you rebase this PR and sort out the conflict that GitHub is reporting please.

lib/Cli.php
```diff
@@ -456,7 +456,7 @@ protected function convert($vObj)
      */
     protected function color($vObj)
     {
-        fwrite($this->stdout, $this->serializeComponent($vObj));
+        $this->serializeComponent($vObj);
```
phil-davis, line 459: I am not familiar with what this `color()` method was supposed to do.
Was it supposed to write output? Or?

lib/Component.php
```diff
@@ -238,7 +238,13 @@ public function select($name)
...
                     $result,
                     function ($child) use ($group) {
-                        return $child instanceof Property && strtoupper($child->group) === $group;
+                        if ($child instanceof Property) {
+                            $cgroup = isset($child->group) ? strtoupper($child->group) : '';
+
+                            return $cgroup === $group;
+                        }
+
+                        return false;
```
phil-davis: yes, this looks a bit difficult to "engineer" a test to exactly cover this line. Actually the previous check `$child instanceof Property` before `&&` was never covered for the `false` case anyway, but Codecov was not fussy about detecting coverage of each boolean true/false alternative in a logical expression.

So we are really in the same position as before - it is just that Codecov now tells us.

mstilkerich:
Thank you. I rebased. The conflicting line was already changed to essentially fix the same issue, but I believe it was not correctly fixed. A filter for "." should match any property without a group according to the description of the function, but with the current master it will not return anything. I also added a test case to show this.

lib/Component.php
phil-davis: This code has been refactored back into a single-line statement.

lib/Cli.php
mstilkerich: Well you tell me :-) It appears this function implements the `bin/vobject color` command which can output a color highlighted version of a vobject file such as a vcard. However, the called serializeComponent() function already writes to stdout by itself, and returns nothing, so the write on void that was previously there just makes no sense.

phil-davis: Ahhh - the return value of `serializeComponent` is void! That had not got through to my brain. Now I understand why the `fwrite()` wrapping this makes no sense.

phil-davis (approved):

mstilkerich:
Thank you!

phil-davis:
@mstilkerich thanks - I will look through other open PRs this evening and see what is an easy-win fix and merge those, then publish a release, hopefully in a few hours.

PR #562, opened 2022-01-21
Modernize code

tcitworld:
* Require PHP ~7.4~ 7.3
* Add typing everywhere possible
* Small internal improvements


staabm:
thanks for the contribution.

this PR is to big for me to review. I can not afford the time to review such big changes.

maybe other maintainers see it differently, but I cannot invest my time here.

phil-davis:
First question - do we want to already drop support for those old PHP versions. PHP 7.3 only just went off the end of support.
Maybe we should keep 7.3 support for another 6 months?

@staabm @DeepDiver1975 

(I will push the button to run the CI, so we at least see the result)

phil-davis:
CI for the older versions is going to fail. @tcitworld you could remove the CI pipelines for 7.1 and 7.2 - assuming that we will be OK to drop those.

tcitworld:
> you could remove the CI pipelines for 7.1 and 7.2

Done.

Sorry for such a big commit but there's no easy way to split things up. I can comment inline on changes that aren't just putting types, if that helps.

phil-davis:
@tcitworld the code-style checker in CI suggests (insists!) on removing the "redundant" PHPdoc of types that are already declared inline in the code.

lib/Parameter.php
```diff
@@ -356,12 +334,12 @@ public function __toString()
...
     #[\ReturnTypeWillChange]
-    public function getIterator()
+    public function getIterator(): ElementList
     {
         if (!is_null($this->iterator)) {
             return $this->iterator;
         }
 
-        return $this->iterator = new ArrayIterator((array) $this->value);
+        return $this->iterator = new ElementList((array) $this->value);
```
tcitworld, line 343: Default might as well be an `ElementList` directly

lib/ITip/Broker.php
```diff
@@ -955,9 +949,7 @@ protected function parseEventInfo(VCalendar $calendar = null)
 
         asort($significantChangeEventProperties);
 
-        foreach ($significantChangeEventProperties as $eventSignificantChangeHash) {
-            $significantChangeHash .= $eventSignificantChangeHash;
-        }
+        $significantChangeHash = implode('', $significantChangeEventProperties);
```
tcitworld, line 952: Simplify (will still return an empty string even with an empty array)

lib/ITip/Broker.php
```diff
@@ -841,7 +835,7 @@ protected function parseEventInfo(VCalendar $calendar = null)
             if (isset($vevent->ORGANIZER)) {
                 if (is_null($organizer)) {
                     $organizer = $vevent->ORGANIZER->getNormalizedValue();
-                    $organizerName = isset($vevent->ORGANIZER['CN']) ? $vevent->ORGANIZER['CN'] : null;
+                    $organizerName = $vevent->ORGANIZER['CN'] ?? null;
```
tcitworld, line 838: Simplify

lib/ITip/Broker.php
```diff
@@ -261,30 +267,23 @@ public function parseEvent($calendar, $userHref, $oldCalendar = null)
...
             // information. We can just remove all existing components
             // and create new ones.
             foreach ($existingObject->getComponents() as $component) {
                 $existingObject->remove($component);
             }
-            foreach ($itipMessage->message->getComponents() as $component) {
-                $existingObject->add(clone $component);
-            }
+        }
+        foreach ($itipMessage->message->getComponents() as $component) {
```
tcitworld, line 285: Extracted common code

lib/Document.php
```diff
@@ -248,17 +231,9 @@ public function getClassNameForPropertyValue($valueParam)
...
      */
-    public function getClassNameForPropertyName($propertyName)
+    public function getClassNameForPropertyName(string $propertyName): string
     {
-        if (isset(static::$propertyMap[$propertyName])) {
-            return static::$propertyMap[$propertyName];
-        } else {
-            return Property\Unknown::class;
-        }
+        return static::$propertyMap[$propertyName] ?? Property\Unknown::class;
```
tcitworld, line 237: Simplify

lib/Document.php
```diff
@@ -97,22 +93,20 @@ public function __construct()
         $args = func_get_args();
         $name = static::$defaultName;
         if (0 === count($args) || is_array($args[0])) {
-            $children = isset($args[0]) ? $args[0] : [];
-            $defaults = isset($args[1]) ? $args[1] : true;
+            $children = $args[0] ?? [];
```
tcitworld, line 96: Simplify

lib/DateTimeParser.php
```diff
@@ -86,80 +81,90 @@ public static function parseDate($date, DateTimeZone $tz = null)
...
-     * @param string $duration
-     * @param bool   $asString
+     * This method will return a DateTimeInterval object
      *
-     * @return DateInterval|string
+     * @throws InvalidDataException
+     * @throws Exception
      */
-    public static function parseDuration($duration, $asString = false)
+    public static function parseDuration(string $duration): DateInterval
```
tcitworld, line 89: In order for the method to return a single type, we split it in two (see `parseDurationAsString` below, which is used only in tests).

lib/Component/VAlarm.php
```diff
@@ -53,17 +60,15 @@ public function getEffectiveTriggerTime()
...
                     $effectiveTrigger = $parentComponent->DTSTART->getDateTime();
                     $duration = VObject\DateTimeParser::parseDuration($parentComponent->DURATION);
                     $effectiveTrigger = $effectiveTrigger->add($duration);
-                    $effectiveTrigger = $effectiveTrigger->add($triggerDuration);
                 } else {
                     $effectiveTrigger = $parentComponent->DTSTART->getDateTime();
-                    $effectiveTrigger = $effectiveTrigger->add($triggerDuration);
                 }
             }
+            $effectiveTrigger = $effectiveTrigger->add($triggerDuration);
```
tcitworld, line 71: Extracted common call

lib/Component.php
```diff
@@ -282,42 +278,39 @@ public function serialize()
...
+                    $score = 100000000;
+                } else {
+                    // All other properties
+                    $score = 200000000;
                 }
+
+                return $score + $key;
             }
+
+            return 0;
```
tcitworld, line 306: Now returning 0 in the last case.

lib/Component.php
```diff
@@ -228,7 +226,7 @@ public function select($name)
         }
 
         if (!is_null($name)) {
-            $result = isset($this->children[$name]) ? $this->children[$name] : [];
+            $result = $this->children[$name] ?? [];
```
tcitworld, line 229: Simplify

lib/Cli.php
```diff
@@ -513,42 +494,39 @@ protected function serializeComponent(Component $vObj)
...
+                    $score = 100000000;
+                } else {
+                    // All other properties
+                    $score = 200000000;
                 }
+
+                return $score + $key;
             }
+
+            return 0;
```
tcitworld, line 522: Now returning `0` in the last case

lib/Cli.php
```diff
@@ -20,13 +25,6 @@ class Cli
      */
     protected $quiet = false;
 
-    /**
-     * Help display.
-     *
-     * @var bool
-     */
-    protected $showHelp = false;
```
tcitworld, line 28: Unused.

composer.json
```diff
@@ -90,7 +91,7 @@
     },
     "scripts": {
         "phpstan": [
-            "phpstan analyse lib tests"
+            "phpstan analyse lib tests --memory-limit 1G"
```
tcitworld, line 94: There's possibly a lower appropriate value.

staabm:
> PHP 7.3 only just went off the end of support.
> Maybe we should keep 7.3 support for another 6 months?

fine with me

lib/Component.php
```diff
@@ -13,6 +17,8 @@
  * @copyright Copyright (C) fruux GmbH (https://fruux.com/)
  * @author Evert Pot (http://evertpot.com/)
  * @license http://sabre.io/license/ Modified BSD License
+ *
+ * @property VObject\Property\FlatText UID
```
tcitworld, line 21: I've added this kind of annotations each time a magic property was been used, to improve IDE DX and possibly static analysis.

phil-davis:
@tcitworld you should be able to locally do:
`composer cs-fixer`

and that will make the code style edits that php-cs-fixer insists on having.

tcitworld:
For some reason php-cs-fixer didn't report anything (running on php 8, might be the issue). Had to do the edits manually.

lib/Cli.php
```diff
@@ -130,7 +126,6 @@ public function main(array $argv)
                         $this->showHelp();
 
                         return 0;
-                        break;
```
tcitworld, line 133: Useless break after return

lib/Cli.php
```diff
@@ -218,7 +213,7 @@ public function main(array $argv)
         $command = $positional[0];
 
         $this->inputPath = $positional[1];
-        $this->outputPath = isset($positional[2]) ? $positional[2] : '-';
+        $this->outputPath = $positional[2] ?? '-';
```
tcitworld, line 216: Simplify

lib/Cli.php
```diff
@@ -451,22 +439,18 @@ protected function convert($vObj)
...
     /**
      * Colorizes a file.
-     *
-     * @param Component $vObj
      */
-    protected function color($vObj)
+    protected function color(Component $vObj): int
     {
-        fwrite($this->stdout, $this->serializeComponent($vObj));
+        $this->serializeComponent($vObj);
```
tcitworld, line 445: `$this->serializeComponent($vObj)` does its `fwrite` itself and returns `void` so it makes no sense to `fwrite` it itself.

phil-davis:
@tcitworld CI passes - that's great.

We want to still support PHP 7.3. For example, both ownCloud and NextCloud use the `sabre` libraries, and they both still support PHP 7.3 (for however many months). And probably there are other consumers of `sabre` that are similar. So we don't want to prevent those from being able to use releases.

Please add back the PHP 7.3 CI, adjust `composer.json` etc for PHP 7.3 and see what fails (if anything)

The PHP 7.4 new features are documented here:
https://www.php.net/manual/en/migration74.new-features.php

I you have already used any of those, then that will need to be reverted (or this PR will get "stuck" for probably 6 months)

tcitworld:
I removed the typed properties and restored type information as phpdoc. Will open another PR with php 7.4 specific features later.

composer.json
```diff
@@ -32,15 +32,16 @@
     "homepage" : "http://sabre.io/vobject/",
     "license" : "BSD-3-Clause",
     "require" : {
-        "php"          : "^7.1 || ^8.0",
+        "php"          : "^7.3 || ^7.4 || ^8.0",
```
tcitworld, line 35:
```suggestion
        "php"          : "^7.3 || ^8.0",
```

composer.json
phil-davis: That's fine - I guess it is just nice to have it "explode" gracefully if some analysis eats up memory, rather than locking up a whole machine when physical memory runs out.

phil-davis:
@tcitworld time has gone by and we had another discussion about this sort of stuff - https://github.com/sabre-io/xml/issues/215
We are going to drop PHP 7.3 and just keep support for PHP 7.4 and up. I did a PR like this in sabre/xml https://github.com/sabre-io/xml/pull/220 which has a much smaller code-base.

I will cherry-pick the code you have provided here into a branch of my own, rebase it and try to get back the PHP 7.4-only things.

Sorry for the pain and long delay, but thanks for the work you have done in the lots of detail needed for this.

tcitworld:
Thanks! You may use [this branch](https://github.com/tcitworld/vobject/tree/modernize-php-7.4) as well that I had kept.

phil-davis:
Note: this will fix #584

phil-davis:
Note: see #586 for the latest code for this. I took https://github.com/tcitworld/vobject/tree/modernize-php-7.4 and rebased it, then kept going myself. IMO that is ready to become a first RC1 for a major release 5.0.0

After that is sorted out and merged, we can close this PR, because the code here will have found its way into master.

phil-davis:
PR #586 has been merged - thanks @tcitworld 
I will do a PR to increase the phpstan level, then we should be able to release a major version 5.

PR #564, opened 2022-02-14
Fix infinite loop caused by yearly with bySetPos

liurxliu:
Consider the rrule:
`FREQ=YEARLY;BYMONTH=5;BYSETPOS=3;BYMONTHDAY=3`

It will not generate any occurrence, but the `RRuleIterator` will stuck in the infinite loop

### DIFF
--- a/lib/Recur/RRuleIterator.php
+++ b/lib/Recur/RRuleIterator.php
@@ -658,6 +658,14 @@ class RRuleIterator implements Iterator
                     (int) $currentMonth,
                     (int) $currentDayOfMonth
                 );
+
+                // To prevent running this forever (better: until we hit the max date of DateTimeImmutable) we simply
+                // stop at 9999-12-31. Looks like the year 10000 problem is not solved in php ....
+                if ($this->currentDate->getTimestamp() > 253402300799) {
+                    $this->currentDate = null;
+
+                    return;
+                }
             }
 
             // If we made it here, it means we got a valid occurrence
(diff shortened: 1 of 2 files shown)


phil-davis:
`FREQ=YEARLY;BYMONTH=5;BYSETPOS=3;BYMONTHDAY=3`

I think that means that the event happens yearly, on the 5th month and the 3rd day - for example 2022-05-03, just one event in a year. But then `BYSETPOS=3` means, of all the events in the year, take only the 3rd event. But there is only ever one event in any year, so the `BYSETPOS=3` filter means that no events ever match the criteria and we get an infinite loop that tries to find a "next" event in the future, but there is never an event.

It seems "a good thing" to "forcibly" stop the infinite loop at some point, to catch all these sorts of cases where the rule iterator will never get a match.

But it would also be nice to be able to analyse/parse rules "up front" and deduce that the rule can, mathematically, never generate any match, and so it is pointless wasting the effort of looping from now to year 9999.

Maybe there are a lot more rule cases that result in there never being a match for some yearly/monthly/weekly... rule? And so there might be infinite loops waiting in monthly/weekly etc processing?

liurxliu:
You are right! It will be more efficient to be able to detect rrule that never generate any occurrence. The fix was took from the monthly case here: https://github.com/sabre-io/vobject/issues/442

phil-davis:
Current CI passes in my test PR #564 so this works OK with the current code base and CI.

Let's take this fix - it fixes the problem in the same way as the existing code added in #443 - so we might as well take the same approach for now. I will merge.

phil-davis (approved):

PR #565, opened 2022-02-20
lowercase mailto:-addresses in getNormalizedValue()

pk1234:
so far getNormalizedValue() only lowercased schema-value.
for mailto:-addresses it now lowercases the address as well.

### DIFF
--- a/lib/Property/ICalendar/CalAddress.php
+++ b/lib/Property/ICalendar/CalAddress.php
@@ -54,7 +54,7 @@ class CalAddress extends Text
         }
         list($schema, $everythingElse) = explode(':', $input, 2);
         $schema = strtolower($schema);
-        if($schema == "mailto") {
+        if($schema === "mailto") {
             $everythingElse = strtolower($everythingElse);
         }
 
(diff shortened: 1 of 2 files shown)


staabm (approved):
Lgtm

lib/Property/ICalendar/CalAddress.php
```diff
@@ -53,7 +53,11 @@ public function getNormalizedValue()
             return $input;
         }
         list($schema, $everythingElse) = explode(':', $input, 2);
+        $schema = strtolower($schema);
+        if($schema == "mailto") {
```
DeepDiver1975, line 57: should be  ===

DeepDiver1975 (dismissed):

lib/Property/ICalendar/CalAddress.php
pk1234: strtolower() returns a string value, so there is no real difference between == and === here.
But you are right. When using a language where "0xFF" == "255", one should always compare strings with the === operator.

phil-davis (approved):

PR #568, opened 2022-04-02
Add NICKNAME fallback for VCards without FNs

Evengard:
This is a rare yet possible case (I stumbled upon such VCards in my own Google Account storage) where a NICKNAME is defined on a VCard but no FN. Importing such contacts into Nextcloud (using Sabre/VObject) is a bit of a pain, this attempts to solve it.

### DIFF
--- a/tests/VObject/Component/VCardTest.php
+++ b/tests/VObject/Component/VCardTest.php
@@ -92,6 +92,14 @@ class VCardTest extends TestCase
             ],
             "BEGIN:VCARD\r\nVERSION:4.0\r\nUID:foo\r\nORG:Acme Co.\r\nFN:Acme Co.\r\nEND:VCARD\r\n",
         ];
+        // No FN, NICKNAME fallback
+        $tests[] = [
+            "BEGIN:VCARD\r\nVERSION:4.0\r\nUID:foo\r\NICKNAME:JohnDoe\r\nEND:VCARD\r\n",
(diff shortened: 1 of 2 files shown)


Evengard:
Is that project not maintained anymore?...

phil-davis (approved):
This seems reasonable. If the Formatted Name "FN" is not provided, then we have a "reasonable try" to fill it in by trying to use "N" (Name), "ORG", "NICKNAME", "email" - whichever is found first in that order.

tests/VObject/Component/VCardTest.php
```diff
@@ -92,6 +92,14 @@ public function validateData()
             ],
             "BEGIN:VCARD\r\nVERSION:4.0\r\nUID:foo\r\nORG:Acme Co.\r\nFN:Acme Co.\r\nEND:VCARD\r\n",
         ];
+        // No FN, NICKNAME fallback
+        $tests[] = [
+            "BEGIN:VCARD\r\nVERSION:4.0\r\nUID:foo\r\NICKNAME:JohnDoe\r\nEND:VCARD\r\n",
```
phil-davis, line 97: Note: CI did not run in this PR. This line was missing an "n"
`UID:foo\r\NICKNAME:JohnDoe` needs to be `UID:foo\r\nNICKNAME:JohnDoe`

Fixed in #576

PR #571, opened 2022-04-29
Fully ignore invalid lines

heiglandreas:
Previously when OPTION_IGNORE_INVALID_LINES was set any invalid content of a property was ignored.

With this change now also invalid properties themselves are ignored.

This will help importing feeds that contain broken multi-line contents where newlines are not propperly escaped and therefore parts of the multiline content show up as property themselves. Without this addition the import will then break and not import anything at all. With this change added an import will now succeed when invalid lines shall be ignored.

Without the option set everything will behave as before.

### DIFF
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -372,6 +372,13 @@ class MimeDir extends Parser
                 $value = $this->unescapeParam($value);
 
                 if (is_null($lastParam)) {
+                    if ($this->options & self::OPTION_IGNORE_INVALID_LINES) {
+                        // When the property can't be matched and the configuration
+                        // option is set to ignore invalid lines, we ignore this line
+                        // This can happen when servers provide faulty data as iCloud
+                        //  frequently does with X-APPLE-STRUCTURED-LOCATION
+                        continue;
+                    }
                     throw new ParseException('Invalid Mimedir file. Line starting at '.$this->startLine.' did not follow iCalendar/vCard conventions');
                 }
                 if (is_null($property['parameters'][$lastParam])) {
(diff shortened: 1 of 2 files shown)


lib/Parser/MimeDir.php
```diff
@@ -372,6 +372,13 @@ protected function readProperty($line)
                 $value = $this->unescapeParam($value);
 
                 if (is_null($lastParam)) {
+                    if ($this->options & self::OPTION_IGNORE_INVALID_LINES) {
+                        // When the property can't be matched and the configuration
+                        // option is set to ignore invalid lines, we ignore this line
+                        // This can happen when servers provide faulty data as iCloud
+                        //  frequently does with X-APPLE-SRTUCTURED-LOCATION
```
phil-davis, line 379:
```suggestion
                        //  frequently does with X-APPLE-STRUCTURED-LOCATION
```

I am not sure if the typo here is intentional - does iCloud have/had a bug where it uses `X-APPLE-SRTUCTURED-LOCATION` (with the typo?

phil-davis: @heiglandreas is this just a typo?

phil-davis:
@heiglandreas I am doing some work in this repo, so I have time to look over the pending PRs.
GitHub is reporting conflicts in MimeDirTest.php - please rebase and resolve the conflicts, and either fix the `X-APPLE-SRTUCTURED-LOCATION` typo or let me know if it is intentional.

phil-davis:
@heiglandreas I am allowed to push to your branch - so I have rebased and fixed the typo.

heiglandreas:
Thanks!

phil-davis (approved):

PR #582, opened 2022-08-17
Fix bug in by year day

phil-davis:
This is a rebase of #436


tests/VObject/Recur/RRuleIteratorTest.php
```diff
@@ -569,13 +586,13 @@ public function testYearlyByYearDayMultiple()
...
-                '2017-10-29 14:53:11',
-                '2020-07-09 14:53:11',
-                '2020-10-28 14:53:11',
+                '2011-10-28 14:53:11',
+                '2014-07-09 14:53:11',
+                '2014-10-28 14:53:11',
+                '2017-07-09 14:53:11',
+                '2017-10-28 14:53:11',
+                '2020-07-08 14:53:11',
+                '2020-10-27 14:53:11',
```
phil-davis, line 595: Note for future readers: 202 was a leap year, so day 60 was 29 Feb. In other years, day 60 is 1 Mar. So day 190 and day 301 in a leap year are 2020-07-08 and 2020-10-27 (a day earlier in those months)

This is correct, and so this test (and some other tests) have cases in the output that verify that leap-year counting is happening.

Issue #584, opened 2022-08-17
vobject 4.5 requires sabre xml 2.1


mstilkerich:
Hello, the new release does not allow combination with the also newly released sabre xml 3. I guess this is not intended.

phil-davis:
I tried in #585 but the code is not quite ready yet in `sabre/vobject` - see my comment https://github.com/sabre-io/vobject/pull/585#issuecomment-1218945269

Also see #562 - I plan to grab that, rebase it, and get it ready today. That will get the type declarations etc all up-to-date.

mstilkerich:
Intended after all then! Thanks for the information.

phil-davis:
> Intended after all then! Thanks for the information.

not so much "intended" as "didn't think hard about it or try it" and it turned out anyway that we need to wait!

PR #585, opened 2022-08-18
Allow sabre/xml major version 3

phil-davis:
Issue #584

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require" : {
         "php"          : "^7.1 || ^8.0",
         "ext-mbstring" : "*",
-        "sabre/xml"    : "^2.1"
+        "sabre/xml"    : "^2.1 || ^3.0"
     },
     "require-dev" : {
         "friendsofphp/php-cs-fixer": "~2.17.1",


phil-davis:
phpstan correctly reports;
```
Fatal error: Declaration of Sabre\VObject\Node::xmlSerialize(Sabre\Xml\Writer $writer) must be compatible with                      
     Sabre\Xml\XmlSerializable::xmlSerialize(Sabre\Xml\Writer $writer): void in /home/phil/git/sabre-io/vobject/lib/Node.php on line 85  
```

phpunit also notices this:
```
PHP Fatal error:  Declaration of Sabre\VObject\Node::xmlSerialize(Sabre\Xml\Writer $writer) must be compatible with Sabre\Xml\XmlSerializable::xmlSerialize(Sabre\Xml\Writer $writer): void in /home/runner/work/vobject/vobject/lib/Node.php on line 85
```

`xmlSerialize` has explicit return type `void` in `sabre/xml` major version 3. Anything that inherits from that and redeclares/reimplements that method has to also declare the return type `void`.

`sabre/vobject` needs various return type declarations sorted out. That is my next task anyway. And that will need a najor version release of vobject. At that time it can allow (and will probably require) `sabre/xml` v3.

PR #586, opened 2022-08-19
PHP 7.4 types

phil-davis:
Add parameter and return types etc. that are available in PHP 7,4 and upwards.

Drop support for PHP older than 7.4

code from @tcitworld plus cs-fixer changes.

I went through the code and added whatever else that I could see for declaring types in PHP 7.4

And went through the test code (see last commit) to add type declarations, and make use of newer assert methods that my IDE suggested now exist.


composer.json
```diff
@@ -32,15 +32,16 @@
     "homepage" : "http://sabre.io/vobject/",
     "license" : "BSD-3-Clause",
     "require" : {
-        "php"          : "^7.1 || ^8.0",
+        "php"          : "^7.4 || ^8.0",
         "ext-mbstring" : "*",
-        "sabre/xml"    : "^2.1"
+        "ext-json"     : "*",
```
phil-davis, line 37: ToDo: check why this needs to be added.

lib/BirthdayCalendarGenerator.php
```diff
@@ -142,9 +141,10 @@ public function getResult()
             }
 
             // Create event.
+            /** @var VEvent $event */
```
staabm, line 144: Do we really need this? Or shouldn't a proper return type for the called method be sufficient?

lib/Cli.php
```diff
@@ -431,9 +418,10 @@ protected function convert($vObj)
         }
 
         if ($forceInput && $vObj->name !== $forceInput) {
-            throw new \Exception('You cannot convert a '.strtolower($vObj->name).' to '.$this->format);
+            throw new Exception('You cannot convert a '.strtolower($vObj->name).' to '.$this->format);
         }
         if ($convertVersion) {
+            /** @var VCard $vObj */
```
staabm, line 424: Same here. Inline `@var` should be double checked..

lib/BirthdayCalendarGenerator.php
phil-davis: I am looking through the diffs at the moment, and quite a few things like this will disappear.
cs-fixer finds and removes a lot of them.

phil-davis:
Most of this code is a rebase of code that @tcitworld has handed over to me. I am looking through to see what else can/needs to be done. That will take "some time" because there is a lot to look at...

composer.json
tcitworld: The JSON parser makes uses of `json_decode`. The JSON extension is pretty much bundled everywhere anyway, so it's more about making things sure (and my IDE happy).

lib/BirthdayCalendarGenerator.php
phil-davis: It is not needed. phpstan does not complain without it, so it is not a helper for code-analysis. I removed it.

lib/Cli.php
phil-davis: removed

lib/DateTimeParser.php
```diff
@@ -86,80 +81,90 @@ public static function parseDate($date, DateTimeZone $tz = null)
...
-     * @param string $duration
-     * @param bool   $asString
+     * This method will return a DateTimeInterval object
      *
-     * @return DateInterval|string
+     * @throws InvalidDataException
+     * @throws Exception
      */
-    public static function parseDuration($duration, $asString = false)
+    public static function parseDuration(string $duration): DateInterval
```
phil-davis, line 89: ToDo: understand why we can remove the `$asString` parameter here. This parameter is never used internally, but people using `sabre/vobject` could call this `Sabre\VObject\DateTimeParser::parseDuration` static function and us the `$asString` parameter.

@tcitworld do you remember why you removed the parameter?

tcitworld: I probably thought it would be nicer to have consistent parameters types and return types. Sounds a sane BC.

phil-davis: I will start making notes for the release notes - need to mention this sort of stuff.

phil-davis:
@staabm @tcitworld IMO this is all good, and as close as I can see to implementing all the PHP 7.4 features throughout the code.

Please let me know what you think (I know that you will not likely to be able to look at every line of the diffs!). If you are OK with this, then I could make a major version 5.0.0-RC1 which would not be an official release, but I (and anyone else) could try it in their projects and give feedback about anything that explodes unexpectedly.

lib/BirthdayCalendarGenerator.php
```diff
@@ -15,23 +17,19 @@ class BirthdayCalendarGenerator
 {
     /**
      * Input objects.
```
staabm, line 19: defining array key/value brings a lot of benefits. in the next PR we maybe should increase the phpstan level, so these types are required

```suggestion
     * Input objects.
     *
     * @var list<Component\VCard>
```

but feel free to do these in a followup as this PR already is pretty big

lib/DateTimeParser.php
staabm: we can do such changes more easily/with confidence when we declare the classes which are not designed to be subclasses as `final` that way.

staabm (approved):
I think we should level up phpstan in a future PR.

.php-cs-fixer.dist.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-$config = PhpCsFixer\Config::create();
+$config = new PhpCsFixer\Config();
 $config->getFinder()
     ->exclude('vendor')
     ->in(__DIR__);
```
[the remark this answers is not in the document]
phil-davis:
```
    ->append([
        __FILE__,
    ]);
```
adding that now.

bin/bench_freebusygenerator.php
```diff
@@ -11,7 +11,7 @@
     echo "The process will be repeated 100 times to get accurate stats\n";
     echo "\n";
     echo 'Usage: '.$argv[0]." inputfile.ics\n";
-    exit();
+    exit;
```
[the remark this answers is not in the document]
phil-davis: done

lib/BirthdayCalendarGenerator.php
phil-davis: yes, I will followup with a next PR...

phil-davis:
Merged - I will followup an increased `phpstan` level in a next PR, before making a 5.0.0-RC1 pre-release that can be easily used to test in projects that use this.

Issue #588, opened 2022-08-29
Release major version 5


phil-davis:
which will support PHP 7.4 8.0 8.1 8.2 and have type declarations on parameters and return values etc as far as possible with features available in PHP 7.4.

If possible, increase the phpstan level to 6. That requires adding a lot of parameters and return type detail in PHPdoc (specially of the shape of arrays. When I do this locally, I get a longggg list of things reported. So it might take a while to sort all that out! I will try adjusting a few things and then report how it goes, and how big a job it seems.

mstilkerich:
Hello, was this accidentally published as 4.5.2? 4.5.2 includes huge changes to 4.5.1 including those mentioned in the description of this issue. It also breaks static analysis as the type declarations on many methods have been changed.

Btw, all the `@property` declarations are wrong. `@property` must include the `$` with the property name, e.g. instead of 
`@property VObject\Property\FlatText FN` 
it should be
`@property VObject\Property\FlatText $FN` 

See
https://docs.phpdoc.org/latest/guide/references/phpdoc/tags/property.html
https://phpstan.org/writing-php-code/phpdocs-basics#magic-properties
https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-seal-properties

phil-davis:
@mstilkerich thanks for pointing this out!

I messed up - I should have cherry-picked only the relevant PR onto the 4.5 branch and released from there. I will do that now, and release a 4.5.3

That should avoid people using the not-yet-ready code that was accidentally released as 4.5.2

phil-davis:
I have released a 4.5.3 from the 4.5 branch. https://github.com/sabre-io/vobject/releases/tag/4.5.3
See my comment there. I think that this should correct the situation, leaving only a couple of very minor added return-type declarations.

Please comment if 4.5.3 causes trouble.

mstilkerich:
Thanks, yep on my side 4.5.3 is good again. I really only had issues with the psalm type analysis, mostly caused by the change on the magic `__get()` method of `Component` that used to have a type annotation of returning a `?Property`, but now can also return a `Component`. And component lacks a ``toString()`` method, which will cause type-analysis issues  in simple uses like `$vcard->FN` in a context where a string is needed, which is one of the beauties of this lib.

[...]

PR #591, opened 2022-09-11
Type declarations to increase phpstan to level 6

phil-davis:
1) Bump dependencies in composer.json

2) Bump phpstan to level 6

3) WIP - fixup type declarations (test code)

4) ToDo - fixup type declarations (real code)

Towards issue #588

### DIFF
--- a/lib/Parameter.php
+++ b/lib/Parameter.php
@@ -19,11 +19,6 @@ use Sabre\Xml;
 class Parameter extends Node
 {
     /**
-     * Parameter name.
-     */
-    public string $name;
-
-    /**
      * vCard 2.1 allows parameters to be encoded without a name.
      *
      * We can deduce the parameter name based on its value.
--- a/lib/Parameter.php
+++ b/lib/Parameter.php
@@ -160,7 +160,7 @@ class Parameter extends Node
      *
      * This may be either a single, or multiple strings in an array.
      *
-     * @param string|array $value
+     * @param int|string|array $value
      */
     public function setValue($value): void
     {
@@ -212,7 +212,7 @@ class Parameter extends Node
      * If the argument is specified as an array, all items will be added to the
      * parameter value list.
      *
-     * @param string|array $part
+     * @param int|string|array $part
      */
     public function addValue($part): void
     {
--- a/tests/VObject/EmClientTest.php
+++ b/tests/VObject/EmClientTest.php
@@ -3,6 +3,9 @@
 namespace Sabre\VObject;
 
 use PHPUnit\Framework\TestCase;
+use Sabre\VObject\Component\VCalendar;
+use Sabre\VObject\Component\VEvent;
+use Sabre\VObject\Property\ICalendar\DateTime;
 
 class EmClientTest extends TestCase
 {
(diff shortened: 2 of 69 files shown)


phil-davis:
"progress" report: phpstan [ERROR] Found 638 errors

There were 880 errors, I have fixed 242 - that is 27.5% done! This is a painful job.

lib/Parameter.php
```diff
@@ -160,7 +160,7 @@ public static function guessParameterNameByValue(string $value): string
      *
      * This may be either a single, or multiple strings in an array.
      *
-     * @param string|array $value
+     * @param int|string|array $value
```
staabm, line 163: is it possible to define more precise array-key and value types?

tests/VObject/EmClientTest.php
```diff
@@ -48,8 +51,13 @@ public function testParseTz(): void
 END:VEVENT
 END:VCALENDAR';
 
+        /** @var VCalendar<int, mixed> $vObject */
```
staabm, line 54: can we define a more precise return-type instead?

https://github.com/sabre-io/vobject/blob/7c585e61435306d4030f7e97260d6292892173ba/lib/Reader.php#L43

`@var` has the problem, that we need to declare it on every call-site

tests/VObject/EmptyParameterTest.php
```diff
@@ -51,6 +53,7 @@ public function testVCard21Parameter(): void
         $vcard = new Component\VCard([], false);
         $vcard->VERSION = '2.1';
         $vcard->PHOTO = 'random_stuff';
+        /* @phpstan-ignore-next-line 'Cannot call method add() on string' */
         $vcard->PHOTO->add(null, 'BASE64');
```
staabm, line 57: how can we call `->add` on a property, which we assigned a `string` shortly before?

is this related to magic `__get` & `__set` or similar, and we should/can use a extension in phpstan for that?
https://phpstan.org/developing-extensions/class-reflection-extensions

lib/Parameter.php
phil-davis: probably - I will look in more detail when I get to do a lot a "real code".

tests/VObject/EmptyParameterTest.php
phil-davis: yes, there is "magic property"  stuff happening, so PHOTO does not really end up being a string!
I haven't dug deep into this. I was not being too concerned about the test code.
But I will probably have to dig into it when I get to the "real" code.

phil-davis:
Types are declared in all test code.
`Found 493 errors`
Nearly half way. And then need to review what happened and try to tidy it all up, be consistent, and provide the "reasonable most" specific type declarations at each point.

phil-davis:
Note to interested people: I have had "other things" happening the last few months. This work here is volunteer work. Hopefully I will get enthusiastic about it again "real soon now" (tm) and have some time to put in.

If anyone else wants to do this stuff, or help with it, feel free to ping me here and we can work out how to move forward.

phil-davis:
Rebased and resolved conflicts. Maybe I can get more work done on this "soon".

phil-davis:
https://github.com/sabre-io/vobject/actions/runs/3985955698/jobs/6833950107
```
Found 389 errors
```

phil-davis:
I rebased, resolved conflicts, applied the latest cs-fixer code-style changes. Now `phpstan` level 6 reports "only":
```
[ERROR] Found 382 errors
```

This looks like a big job.

DeepDiver1975:
I tried to address some stuff here as well
https://github.com/sabre-io/vobject/pull/628/commits

PR #593, opened 2022-09-23
Add support for COMMENT property

miaulalala:
RFC5546 allows for COMMENT properties on all iTIP methods. This PR attempts to extend the current implementation to suppot COMMENTs on REQUEST, REPLY and CANCEL methods currently implemented in the Broker.

Signed-off-by: Anna Larch <anna@nextcloud.com>

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -78,6 +78,7 @@ class Broker
         'RDATE',
         'EXDATE',
         'STATUS',
+        'COMMENT'
     ];
 
     /**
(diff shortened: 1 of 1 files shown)


lib/ITip/Broker.php
```diff
@@ -340,6 +343,9 @@ protected function processMessageReply(Message $itipMessage, ?VCalendar $existin
                 $requestStatus = $vevent->{'REQUEST-STATUS'}->getValue();
                 list($requestStatus) = explode(';', $requestStatus);
             }
+            if(!empty($vevent->{'COMMENT'})) {
+                $comments = $vevent->{'COMMENT'};
```
miaulalala, line 347: get the value? or use clone?

miaulalala:
Closing this as we're going to go a different route with comments

PR #602, opened 2022-12-25
Fix unescaping of URIs in VCard4 and mime type of images in converter

mstilkerich:
This PR fixes two issues encountered on a V4 VCard.

1. RFC6350 allows escaping of semicolons in property values, but vobject did not unescape them in input. data URIs used for inline images in v4 VCard contain both semicolon and comma. When the semicolon is escaped (which is optional for the producer of the VCard), it must be unescaped by the consumer or the resulting URI will still contain the backslash and be malformed.
2. Mime types are case insensitive (RFC2045). In the vcard converter in conversion from v4, the mime type is checked in a case sensitive manner and would thus fail to convert the mime type if it is not lowercase in the input. Changed the handling to treat this case-insensitive.

Tests adapted / extended accordingly.

### DIFF
--- a/tests/VObject/VCardConverterTest.php
+++ b/tests/VObject/VCardConverterTest.php
@@ -409,7 +409,14 @@ OUT;
             $vcard
         );
 
-        $input = $output;
+        $input = <<<IN
+BEGIN:VCARD
+VERSION:3.0
+X-ADDRESSBOOKSERVER-KIND:INDIVIDUAL
+END:VCARD
+
+IN;
+
         $output = <<<OUT
 BEGIN:VCARD
 VERSION:4.0
--- a/tests/VObject/VCardConverterTest.php
+++ b/tests/VObject/VCardConverterTest.php
@@ -38,9 +38,9 @@ FN:Steve
 TEL;PREF=1;TYPE=HOME:+1 555 666 777
 ITEM1.TEL:+1 444 555 666
 ITEM1.X-ABLABEL:CustomLabel
-PHOTO;TYPE=HOME:data:image/jpeg;base64,Zm9v
-PHOTO:data:image/gif;base64,Zm9v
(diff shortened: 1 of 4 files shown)


mstilkerich:
Added another fix: Remove `X-ADDRESSBOOKSERVER-KIND=individual` property in converting vcard v3 to v4. The existing test for this was broken because the used input did not contain the corresponding property.

tests/VObject/VCardConverterTest.php
```diff
@@ -38,9 +38,9 @@ public function testConvert30to40(): void
 TEL;PREF=1;TYPE=HOME:+1 555 666 777
 ITEM1.TEL:+1 444 555 666
 ITEM1.X-ABLABEL:CustomLabel
-PHOTO;TYPE=HOME:data:image/jpeg;base64,Zm9v
-PHOTO:data:image/gif;base64,Zm9v
-PHOTO;X-PARAM=FOO:data:image/png;base64,Zm9v
+PHOTO;TYPE=HOME:data:image/jpeg;base64\\,Zm9v
+PHOTO:data:image/gif;base64\\,Zm9v
```
staabm, line 42: I think we should not change the existing test but add a new one so we keep backwards compatibility...?

staabm: I don't know whether existing clients might also send malformed requests?

Or existing scripts relied on this example

mstilkerich: Just so there is no confusion: vobject will continue to parse an unescaped comma here as before, this is just a change in the test data. If you prefer to keep it malformed, no problem. I thought it would be better if the actual test samples are properly escaped as to not confuse the reader. Escaping of the commas in the test data I did only for this reason and is not required concerning my code changes.

mstilkerich: Hello, will it help to get this merged if I revert all unnecessary changes in the tests?

PR #603, opened 2023-01-19
Support sabre/xml v4

gharlan:
I would like to use sabre/vobject together with latest sabre/xml v4.
Latest vobject release still uses sabre/xml v2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "php"          : "^7.4 || ^8.0",
         "ext-mbstring" : "*",
         "ext-json"     : "*",
-        "sabre/xml"    : "^3.0"
+        "sabre/xml"    : "^3.0 || ^4.0"
     },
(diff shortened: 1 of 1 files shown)


staabm (approved):
lgtm. the main diff of 4.x was the sabre/uri difference described in https://github.com/sabre-io/xml/releases/tag/4.0.0

I think we only need to make sure, when we release this change that we also note about this sabre/uri implication in the releases notes.

the failling builds are interessting though

phil-davis:
See #604 - there are new problems in CI that are unrelated to this PR.
I will see if they can be resolved first, before making the change to sabre/xml v4.

phil-davis (approved):

phil-davis:
https://github.com/sabre-io/vobject/releases/tag/4.5.2 is available.

PR #604, opened 2023-01-20
Adjust code style to meet current php-cs-fixer

phil-davis:
php-cs-fixer is reporting these. I guess there has been a newer php-cs-fixer version released that is different.

It wants "root" classes and functions to use the leading `\` - `\InvalidArgumentException` `\substr`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
         uses: shivammathur/setup-php@v2 #https://github.com/shivammathur/setup-php
         with:
           php-version: ${{ matrix.php-versions }}
+          update: false
           extensions: mbstring, dom, fileinfo, mysql, redis, opcache
           coverage: ${{ matrix.coverage }}
           tools: composer
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


phil-davis:
The php unit test fails for PHP 8.1 and 8.2 are because of https://github.com/php/php-src/issues/10218
Timezones GMT+n were broken in PHP 8.1.14 and 8.2.1 released in early January.

Fixed in commit https://github.com/php/php-src/commit/d12ba111e03fa4e6d76a66a029f5dbe6265b1ea9
But that will likely not be released until Feb 2023.

phil-davis:
https://github.com/sabre-io/vobject/actions/runs/3964660146/jobs/6793716785

It still uses PHP 8.1.14 in the workflow.

I raised issue https://github.com/shivammathur/setup-php/issues/687 to see if there is some way to successfully select an old PHP patch release.

phil-davis:
See #605 for the code that fixes CI.

I will leave this PR here for now, as it demonstrates https://github.com/shivammathur/setup-php/issues/687

phil-davis:
Issue https://github.com/shivammathur/setup-php/issues/687
Specifying an old patch release of PHP is not supported.
To work around the problem/regression in PHP 8.1.14 and 8.2.1 I did:
https://github.com/sabre-io/vobject/commit/3bef1cfb8540f93d35b9d9deb30c033d51c55604

So this PR is no longer needed, closing.

PR #607, opened 2023-01-22
Support sabre/xml v4

phil-davis:
I accidentally released 4.5.2 on Friday with all the code in `master` branch. That code has various PHP 7.4-only things that will not work properly on PHP 7.1 7.2 7.3 etc. 

See comment https://github.com/sabre-io/vobject/issues/588#issuecomment-1398789007 which reports this mistake.

This PR:
1) backports the "Support sabre/xml v4" change to 4.5 branch
2) backports "Skip GMT+n timezone test cases on PHP 8.1.14 and 8.2.1" so that unit tests can be green
3) adds return types to things that implement xmlSerialize
4) adds return types to things that implement xmlDeserialize
5) backports "Prepare 4.5.2" changelog and Version.php change
6) adds changelog and version bump to 4.5.3

If I merge this and then release 4.5.3 from the 4.5 branch, I think that it should be fine.

The added return types are all supported from PHP 7.1 upward, so that is nice.

The only thought that I have now is that if consumers have implemented child classes that redeclare `xmlSerialize` or `xmlDeserialize` then maybe they will also have to add a return type `void` or `array` ? And so that will mean that the release is not strictly backward-compatible. And so a patch release is not correct.

Thoughts please @staabm @DeepDiver1975 or anyone.


staabm:
Alternatively I think we could  release 4.5.1 as 4.5.3 again.

Then we could use 4.5.2 and publish it as a new major.

staabm:
The changes here look fine if you prefer it

phil-davis:
> The changes here look fine if you prefer it

Do you think that the `void` and `array` return type declarations are a "significant" backward-compatibility problem?

If I just re=release 4.5.1 as 4.5.3 then the ability to use `sabre/xml` v3 or v4 (PR #603 ) get to happen.

If I do this PR, then @gharlan (and others) can have the latest `sabre/xml` now.

staabm:
I don't have a strong opinion which way to go from here

PR #609, opened 2023-02-06
Add dependabot to create GH action version update PRs

Rotzbua:
Currently there is no package manager or similar for GH Actions to manage the dependencies.
Dependabot offers to monitor the workflow scripts and propose new versions as PR.
Since GH Actions rarely change and no resources should be wasted, a repetition of one month is sufficient.

Automates chore like  #608

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@
 
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
thanks

PR #616, opened 2023-04-07
Fix Alarm related to DTSTART of VTODO

tpayen:
Following RFC 5545 (https://www.rfc-editor.org/rfc/rfc5545#page-74) the start of a VTODO is DTSTART not DUE
Fixing the tests in consequence

### DIFF
--- a/lib/Component/VAlarm.php
+++ b/lib/Component/VAlarm.php
@@ -39,13 +39,7 @@ class VAlarm extends VObject\Component
             /** @var VEvent|VTodo $parentComponent */
             $parentComponent = $this->parent;
             if ('START' === $related) {
-                if ('VTODO' === $parentComponent->name) {
-                    $propName = 'DUE';
-                } else {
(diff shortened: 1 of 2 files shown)


jarikmarwede:
I have just noticed the same problem to be the cause of an error in my application and have come to the same conclusion for fixing it. 
Is there a reason this isn't merged yet, or can I somehow help in getting this merged? @evert 
Sorry, if the mention isn't correct. You are the one marked as the author in the PHPDoc comment for the `VAlarm` class.

jarikmarwede:
It also seems to me like this is a duplicate of #370, although that PR isn't merged either because it was missing tests in 2017 and now has merge conflicts.

phil-davis:
There are existing cs-fixer and/or phpstan things in master that are causing CI to fail.
I will fix those in PR #667 and then rebase again here.

phil-davis (approved):

Issue #467, opened 2019-09-06
Undefined index: plusminus


mschering:
I encountered an Invite from exchange server that contains this:

BEGIN:VALARM
DESCRIPTION:REMINDER
TRIGGER;RELATED=START:P
ACTION:DISPLAY
END:VALARM

It has a strange trigger value. This gives the following error in sabre/vobject:

feedback": "Undefined index: plusminus",
    "exceptionCode": 0,
    "exceptionClass": "ErrorException",
    "trace": [
        "#0 /usr/local/share/src/www/vendor/sabre/vobject/lib/DateTimeParser.php(107): go\\core\\ErrorHandler::errorHandler(8, 'Undefined index...', '/usr/local/shar...', 107, Array)",
        "#1 /usr/local/share/src/www/vendor/sabre/vobject/lib/Component/VAlarm.php(32): Sabre\\VObject\\DateTimeParser::parseDuration(Object(Sabre\\VObject\\Property\\ICalendar\\Duration))",
        "#2 /usr/local/share/src/www/modules/calendar/model/Event.php(1871): Sabre\\VObject\\Component\\VAlarm->getEffectiveTriggerTime()",

Perhaps this can be ignored?

phil-davis:
Fixed by PR #619 - it will be in the next release.

PR #619, opened 2023-06-20
fixed issue: Undefined index: plusminus #467

sash04ek:
https://github.com/sabre-io/vobject/issues/467

PR #622, opened 2023-07-19
Create Issue467Test.php

sash04ek:
added test for the https://github.com/sabre-io/vobject/issues/467 issue


phil-davis:
I added a small test to PR #619 which demonstrates the problem and passes with the fix.

@DeepDiver1975 @staabm do we want a higher-level test like the one in this PR?

DeepDiver1975:
I think both tests are valid as they operate on different levels.

staabm (approved):

PR #623, opened 2023-08-08
docs: fix type definition for oldCalendar

kesselb:
The type definition for oldCalendar is fixed on master.

I would be happy about a back port for 4.5 to reduce the false-positive warnings by static code analysis.


phil-davis (approved):
LGTM

PR #628, opened 2023-11-08
Phil davis issue 588

DeepDiver1975:
rebased #591


lib/Node.php
```diff
@@ -11,6 +11,8 @@
  * @author Evert Pot (http://evertpot.com/)
  * @license http://sabre.io/license/ Modified BSD License
  */
+
+/** @implements \IteratorAggregate<int, Node> */
```
DeepDiver1975, line 15: I am not too sure if this is the correct approach ... it can also hold sub classes of Node .....

PR #634, opened 2024-01-15
rename for deprecated timezones

JonBendtsen:
rename for deprecated timezones

https://data.iana.org/time-zones/tzdb/NEWS
https://www.php.net/manual/en/timezones.others.php


phil-davis:
@JonBendtsen one unit test fails, something related to timezones.
Please have a look. Maybe the test needs to change, or maybe the updated timezone data needs a small tweak?

JonBendtsen:
> @JonBendtsen one unit test fails, something related to timezones. Please have a look. Maybe the test needs to change, or maybe the updated timezone data needs a small tweak?

@phil-davis 
I have now looked at the reports from PHPUnit for all of the different PHP versions and I can not in that output see anything that indicates precisely where it goes wrong, so I can not even see where I should make a small tweak.

lib/timezonedata/php-bc.php
```diff
@@ -16,67 +16,319 @@
...
-    'Australia/Victoria',
-    'Australia/West',
-    'Australia/Yancowinna',
-    'Brazil/Acre',
-    'Brazil/DeNoronha',
-    'Brazil/East',
-    'Brazil/West',
-    'Canada/Atlantic',
-    'Canada/Central',
-    'Canada/Eastern',
```
phil-davis, line 71: The failing unit test happens to use `Canada/Eastern`, which is removed here.
https://www.php.net/manual/en/timezones.others.php
"Warning: Please do not use any of the timezones listed here (besides UTC), they only exist for backward compatible reasons, and may expose erroneous behavior. Furthermore, these timezones may be removed from the IANA timezone database at any time."

And `Canada/Eastern` is listed there.

The unit test can be fixed by using a "proper" timezone in the test. But we need to think and properly understand what happens in genera if we remove these timezones from the data here.

Issue #635, opened 2024-01-17
createProperty: check if parameter VALUE is a string


martijnvankekem:
In document.php, function `createProperty()`:

For some iCal structures, $parameters["VALUE"] can be of type array instead of type string, resulting in an error when trying to parse it to strtoupper().

To fix this, besides checking if $parameters["VALUE"] is set, we should check if it is a string as well.

PR #636, opened 2024-01-17
Check for string-type parameter VALUE

martijnvankekem:
See: https://github.com/sabre-io/vobject/issues/635

### DIFF
--- a/lib/Document.php
+++ b/lib/Document.php
@@ -190,7 +190,7 @@ abstract class Document extends Component
 
         if (is_null($class)) {
             // If a VALUE parameter is supplied, we should use that.
-            if (isset($parameters['VALUE'])) {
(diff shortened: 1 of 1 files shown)


staabm:
please add a test case

phil-davis:
@martijnvankekem your PR comes directly from the `master` branch of your fork, which is a bit more diffcult to manage.
I rebased in a branch in my fork and made and up-to-date PR #640 
The CI all passes there - good, no existing test cases were broken.

Please let us know the details of an example iCal structure - what does it pass in `$parameters["VALUE"]`
It will be good to add a unite test case that has a reasonably realistic example of this.

phil-davis:
The bit of code that is enhanced here, was removed in https://github.com/sabre-io/vobject/pull/674/changes

So there is nothing for this edit to apply to any more.

If something still needs to be changed, please advise.

PR #640, opened 2024-02-11
Check for string-type parameter VALUE

phil-davis:
Rebase of #636

### DIFF
--- a/lib/Document.php
+++ b/lib/Document.php
@@ -188,7 +188,7 @@ abstract class Document extends Component
 
         if (is_null($class)) {
             // If a VALUE parameter is supplied, we should use that.
-            if (isset($parameters['VALUE'])) {
+            if (isset($parameters['VALUE']) && is_string($parameters['VALUE'])) {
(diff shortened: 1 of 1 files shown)


phil-davis:
CI passes, so existing test cases are still OK.

Need a response to https://github.com/sabre-io/vobject/pull/636#issuecomment-1937466796 to be able to add a useful extra unit test case.

lib/Document.php
```diff
@@ -188,7 +188,7 @@ public function createProperty(string $name, $value = null, ?array $parameters =
 
         if (is_null($class)) {
             // If a VALUE parameter is supplied, we should use that.
-            if (isset($parameters['VALUE'])) {
+            if (isset($parameters['VALUE']) && is_string($parameters['VALUE'])) {
```
phil-davis, line 191: This bit of code was removed in https://github.com/sabre-io/vobject/pull/674/changes

So there is nothing for this edit to apply to any more.

PR #644, opened 2024-04-18
chore: apply cs-fixer 3.54.0 changes

phil-davis:
When I do:
```
composer update
composer cs-fixer
```
it now has PHP-cs-fixer 3.54.0, which suggests these format changes.


phil-davis:
It's a bit annoying to have so many lines of indent changed. But actually the newer release(s) of php-cs-fixer has found things that were not correctly indented.

PR #647, opened 2024-04-22
Reproduce bug where dst leap is passed on to subsequent occurences

schreven:
Part of issue #648

### DIFF
--- a/tests/VObject/Recur/RRuleIteratorTest.php
+++ b/tests/VObject/Recur/RRuleIteratorTest.php
@@ -162,6 +162,64 @@ class RRuleIteratorTest extends TestCase
         );
     }
 
+     /**
+     * @dataProvider dstTransitionProvider
+     */
+    public function testDailyOnDstTransition(string $start, array $expected): void
+    {
+        $this->parse(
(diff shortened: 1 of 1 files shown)


phil-davis:
@schreven thanks for the test cases.
So the problem is that if a requested event time is in the region 0200 <= time < 0300 then that time does not exist on the morning that "summer time" starts. There is 0159 then straight to 0300 "on the clock".
The event for that day does a reasonable thing. For example, a request for a meeting at 0215 gets scheduled at 0315.

But then that daily event gets scheduled at 0315 on the following days, which is not correct. The event should happen at 0215 (now on the "summer time" clock).

Can you raise an issue about this please? (or link to an existing issue)

Do you have a suggested fix?

schreven:
Hey @phil-davis, yes you summarized the issue perfectly. 

I see that you spotted the issue I created https://github.com/sabre-io/vobject/issues/648 and I continued the discussion there. 

Thank you for this library, it's a nice piece of software

phil-davis:
This code, and the fixes related to it was done in #653

Issue #646, opened 2024-04-22
parser error when DATE is empty


JohnRDOrazio:
I was attempting to parse a defective ICS file, which happened to have empty `CREATED:` and `LAST-MODIFIED:` fields:
```ical
BEGIN:VEVENT
CLASS:PUBLIC
DTSTART;VALUE=DATE:20291231
DTSTAMP:20240422T062620Z
UID:c0e7126a841a70761e4057e58b5e2d52
CREATED:
DESCRIPTION:\nA fun event\nRSVP
LAST-MODIFIED:
```

I wasn't getting parsing errors from the `validate()` call, instead I was getting PHP warnings about an uninitialized string offset 0 in the `DateTimeParser::parse()` method, specifically from the line:
```php
if ('P' === $date[0] || ('-' === $date[0] && 'P' === $date[1])) {
```

The PHP warning made me realize that `$date` was empty, and in fact upon inspecting the ICS file I found the empty `CREATED:` and `LAST-MODIFIED:` fields.

Perhaps the parser should catch those situations in which the expected field has an empty value, and throw a catchable error?

https://github.com/sabre-io/vobject/blob/caa7af8e3304c85c63cdb6464ecafe09bb9918f6/lib/DateTimeParser.php#L202-L211

phil-davis:
@JohnRDOrazio can you propose some code, and add a unit test, in a PR.
Then I can review it.

It will be good to handle defective input data, rather than just have PHP "crash".

JohnRDOrazio:
I'm seeing that `MimeDir` does actually keep track of the line index when reading and parsing a file:

https://github.com/sabre-io/vobject/blob/caa7af8e3304c85c63cdb6464ecafe09bb9918f6/lib/Parser/MimeDir.php#L289

https://github.com/sabre-io/vobject/blob/caa7af8e3304c85c63cdb6464ecafe09bb9918f6/lib/Parser/MimeDir.php#L296-L298

Perhaps the `Property` class could have an extra property `public int $lineIndex;` which could be set when creating a new Property? That way each node could keep track of it's own line number within the original file?

JohnRDOrazio:
ok so this is a little bit strange... I've tried adding `$lineIndex` and `$lineString` public properties to the `Property` class:
```php
    public function __construct(Component $root, $name, $value = null, array $parameters = [], $group = null, $lineIndex = null, $lineString = null)
    {
        $this->name = $name;
        $this->group = $group;

        $this->root = $root;

[...]

PR #649, opened 2024-04-23
fix issue #646

JohnRDOrazio:
add `lineIndex` and `lineString` properties to a `Property` Node

issue #646

### DIFF
--- a/tests/VObject/Component/VCalendarTest.php
+++ b/tests/VObject/Component/VCalendarTest.php
@@ -777,23 +777,23 @@ ICS;
         $vcal = VObject\Reader::read($defectiveInput);
         $result = $vcal->validate();
         $warningMessages = [];
-        foreach( $result as $error ) {
+        foreach ($result as $error) {
             $warningMessages[] = $error['message'];
         }
-        self::assertCount(2, $result, 'We expected exactly 2 validation messages, instead we got ' . count( $result ) . ' results:' . implode(', ', $warningMessages) );
-        foreach( $result as $idx => $warning ) {
-            self::assertArrayHasKey( 'node', $warning, 'The validation errors should contain a node key' );
-            self::assertInstanceOf( VObject\Property\ICalendar\DateTime::class, $warning[ 'node' ], 'We expected the defective node to be of type Sabre\VObject\Property\ICalendar\DateTime, instead we got type ' . gettype( $warning['node'] ) );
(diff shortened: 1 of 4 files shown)


JohnRDOrazio:
if this PR is accepted, perhaps documentation could be updated to explain how to access the properties of a node, and warn against simply trying to `json_encode` the resulting errors...

JohnRDOrazio:
I added a Unit Test that will check for `lineIndex` and `lineString` properties when validating defective calendar data.

phil-davis:
@JohnRDOrazio php-cs-fixer wants some formatting adjusted.
You can:
`composer cs-fixer`
and commit and push the changes it makes.

phil-davis (approved):
LGTM - this implements error handling features similar to what happens elsewhere (see the PR and issue #646 discussions.

tests/VObject/Component/VCalendarTest.php
```diff
@@ -756,6 +756,49 @@ public function testCalDAVMETHOD(): void
...
+            switch ($idx) {
+                case 0:
+                    self::assertEquals('10', $warning['node']->lineIndex, 'We expected the "lineIndex" property of the first defective node to be 10, instead it was '.$warning['node']->lineIndex);
+                    self::assertEquals('CREATED:', $warning['node']->lineString, 'We expected the "lineString" property of the first defective node to be "CREATED:", instead it was '.$warning['node']->lineString);
+                    break;
+                case 1:
+                    self::assertEquals('11', $warning['node']->lineIndex, 'We expected the "lineIndex" property of the second defective node to be 11, instead it was '.$warning['node']->lineIndex);
+                    self::assertEquals('LAST-MODIFIED:', $warning['node']->lineString, 'We expected the "lineString" property of the second defective node to be "LAST-MODIFIED:", instead it was '.$warning['node']->lineString);
+                    break;
+            }
```
staabm, line 798: I think most of these error descriptions here do not add information over the default assertion error which would happen.

IMO the added description hurt readability of the test and we should have the descriptions (last arg to the  `assert*` calls) only where it actually adds value.

staabm (approved):
I like the change!

tests/VObject/Component/VCalendarTest.php
JohnRDOrazio: Ok I removed most of the error descriptions

staabm (approved):
thank you!

PR #360, opened 2016-12-21
358 Bugfix Broker::processMessageReply

strokyl:
Before this commit processMessageReply did not work for exceptions with
a reccurent id in an other timezone than the dtstart of the master event.

PR #651, opened 2024-05-08
358 Bugfix Broker::processMessageReply

phil-davis:
Before this commit processMessageReply did not work for exceptions with a reccurent id in an other timezone than the dtstart of the master event.

This is a rebased version of PR #360

### DIFF
--- a/tests/VObject/ITip/BrokerProcessReplyTest.php
+++ b/tests/VObject/ITip/BrokerProcessReplyTest.php
@@ -373,6 +373,129 @@ ICS;
         $this->process($itip, $old, $expected);
     }
 
+    public function testReplyNewExceptionReccurenceIdInDifferentTz(): void
+    {
+        // This is a reply to 1 instance of a recurring event. This should
+        // automatically create an exception.
+        $itip = <<<ICS
+BEGIN:VCALENDAR
+VERSION:2.0
+METHOD:REPLY
+BEGIN:VEVENT
+ATTENDEE;PARTSTAT=ACCEPTED:mailto:foo@example.org
+ORGANIZER:mailto:bar@example.org
+SEQUENCE:2
+RECURRENCE-ID;TZID=Asia/Ho_Chi_Minh:20140725T110000
+UID:foobar
+END:VEVENT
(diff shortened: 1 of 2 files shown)


tests/VObject/ITip/BrokerProcessReplyTest.php
```diff
@@ -373,6 +373,129 @@ public function testReplyNewExceptionTz(): void
         $this->process($itip, $old, $expected);
     }
 
+    public function testReplyNewExceptionReccurenceIdInDifferentTz(): void
+    {
+        // This is a reply to 1 instance of a recurring event. This should
+        // automatically create an exception.
```
phil-davis, line 379: I don't understand why this comment has "This should automatically create an exception."

The source code now handles this situation. 
`TZID=Asia/Ho_Chi_Minh:20140725T110000` is the event recurrence of `DTSTART;TZID=America/Toronto:20140724T000000` that happens the next day, 20140725.

`America/Toronto` is UTC-4, `Asia/Ho_Chi_Minh` is UTC+7. So they are 11 hours apart. `0000` in Toronto is the same time as `1100` in Ho Chi Minh City.

tests/VObject/ITip/BrokerProcessReplyTest.php
```diff
@@ -373,6 +373,129 @@ public function testReplyNewExceptionTz(): void
...
+        $this->process($itip, $old, $expected);
+    }
+
+    /**
+     * @group failing
+     */
+    public function testReplyNewExceptionReccurenceIdInUTC(): void
+    {
+        // This is a reply to 1 instance of a recurring event. This should
+        // automatically create an exception.
```
phil-davis, line 442: I don't understand why this comment has "This should automatically create an exception."

Similar to my other comment above, `America/Toronto` is UTC-4. So `0000` in Toronto is the same as `040000Z`

I don't expect any exception to happen.

phil-davis:
I see that @gharlan has made #652 - I will have a look at that...

phil-davis:
Closing in favor of #652 - IMO that is a cleaner solution.

PR #652, opened 2024-05-08
ITip\Broker: handle timezones in replies to exception events

### LINKED ISSUE #358: Broker::processMessageReply
Currently, `Broker::processMessageReply` ignore exceptions that have a recurrent id in a different time-zone than the DTSTART of the master event.
This cause issue in Sabre when a attendee update his partstat on an instance using a client that use UTC for recurrent id.

gharlan:
fixes #358, closes #360

While rebasing #360, I noticed that it only fixes replies to new exception events. I use a different approach now, which also fixes replies to existing exception events.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -333,6 +333,9 @@ class Broker
 
         // Finding all the instances the attendee replied to.
         foreach ($itipMessage->message->VEVENT as $vevent) {
+            // Use the Unix timestamp returned by getTimestamp as a unique identifier for the recurrence.
+            // The Unix timestamp will be the same for an event, even if the reply from the attendee
+            // used a different format/timezone to express the event date-time.
             $recurId = isset($vevent->{'RECURRENCE-ID'}) ? $vevent->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() : 'master';
             $attendee = $vevent->ATTENDEE;
             $instances[$recurId] = $attendee['PARTSTAT']->getValue();
@@ -346,6 +349,7 @@ class Broker
         // all the instances where we have a reply for.
         $masterObject = null;
         foreach ($existingObject->VEVENT as $vevent) {
+            // Use the Unix timestamp returned by getTimestamp as a unique identifier for the recurrence.
             $recurId = isset($vevent->{'RECURRENCE-ID'}) ? $vevent->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() : 'master';
             if ('master' === $recurId) {
                 $masterObject = $vevent;
@@ -393,6 +397,9 @@ class Broker
                 $newObject = $recurrenceIterator->getEventObject();
                 $recurrenceIterator->next();
 
+                // Compare the Unix timestamp returned by getTimestamp with the previously calculated timestamp.
+                // If they are the same, then this is a matching recurrence, even though its date-time may have
+                // been expressed in a different format/timezone.
                 if (isset($newObject->{'RECURRENCE-ID'}) && $newObject->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() === $recurId) {
                     $found = true;
                 }
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -333,7 +333,7 @@ class Broker
 
         // Finding all the instances the attendee replied to.
         foreach ($itipMessage->message->VEVENT as $vevent) {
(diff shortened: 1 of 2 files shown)


tests/VObject/ITip/BrokerProcessReplyTest.php
```diff
@@ -253,6 +253,73 @@ public function testReplyPartyCrasher(): void
         $this->process($itip, $old, $expected);
     }
 
+    public function testReplyExistingExceptionRecurrenceIdInUTC(): void
```
phil-davis, line 256: Is this the same test case as `testReplyNewExceptionRecurrenceIdInUTC` further down?

tests/VObject/ITip/BrokerProcessReplyTest.php
```diff
@@ -373,6 +440,126 @@ public function testReplyNewExceptionTz(): void
         $this->process($itip, $old, $expected);
     }
 
+    public function testReplyNewExceptionRecurrenceIdInDifferentTz(): void
+    {
+        // This is a reply to 1 instance of a recurring event. This should
+        // automatically create an exception.
```
phil-davis, line 446:
```suggestion
        // This is a reply to 1 instance of a recurring event.
```

I don't understand why this comment had "This should automatically create an exception."

tests/VObject/ITip/BrokerProcessReplyTest.php
```diff
@@ -373,6 +440,126 @@ public function testReplyNewExceptionTz(): void
...
+END:VCALENDAR
+ICS;
+
+        $this->process($itip, $old, $expected);
+    }
+
+    public function testReplyNewExceptionRecurrenceIdInUTC(): void
+    {
+        // This is a reply to 1 instance of a recurring event. This should
+        // automatically create an exception.
```
phil-davis, line 506:
```suggestion
        // This is a reply to 1 instance of a recurring event.
```

I don't understand why this comment had "This should automatically create an exception."

tests/VObject/ITip/BrokerProcessReplyTest.php
gharlan: No, here the exception event already exists in `$old`. So it is a reply to an existing exception event, which wasn't covered/fixed by the original pull request.

lib/ITip/Broker.php
```diff
@@ -333,7 +333,7 @@ protected function processMessageReply(Message $itipMessage, ?VCalendar $existin
 
         // Finding all the instances the attendee replied to.
         foreach ($itipMessage->message->VEVENT as $vevent) {
-            $recurId = isset($vevent->{'RECURRENCE-ID'}) ? $vevent->{'RECURRENCE-ID'}->getValue() : 'master';
+            $recurId = isset($vevent->{'RECURRENCE-ID'}) ? $vevent->{'RECURRENCE-ID'}->getDateTime()->getTimestamp() : 'master';
```
phil-davis, line 336: Note for myself ( and maybe helps others):
https://www.php.net/manual/en/datetime.gettimestamp.php returns the Unix timestamp (seconds since "Unix zero time").

So that is a unique number that corresponds to the day-and-time of the particular event. If that timestamp of a recurrence matches the timestamp of a reply, then we know that it is a reply to that occurrence of the event - seems good to me.

tests/VObject/ITip/BrokerProcessReplyTest.php
gharlan: It is a copy/paste from all existing `testReplyNewException...` test cases.
The original event didn't had the exception, but the attendee replies to one recurrence only. So the reply should create the exception event.

lib/ITip/Broker.php
gharlan: Yes, the unix timestamp avoids the timezone problem

tests/VObject/ITip/BrokerProcessReplyTest.php
phil-davis: My brain was thinking "exception" as in "throw an exception"!
But in this context, "exception event" means a recurrence of a recurring event that has "something different" about it. In this case, it has an attendee who is attending that particular recurrence of the event.

All good.

tests/VObject/ITip/BrokerProcessReplyTest.php
phil-davis: See comments for the same comment in the previous test case.

staabm (approved):
I don't see obvious problems

lib/ITip/Broker.php
phil-davis: I pushed a commit that adds comments about the use of getTimestamp()

phil-davis (approved):

Issue #657, opened 2024-05-23
Fatal error when passing NULL to the Reader


alecpl:
`Reader::read()`'s 1st argument is defined as `string|resource` indeed. However, it accepts NULL, and also `Parser/MimeDir::parse()`'s 1st argument is described as `string|resource|null`. So, there's some inconsistency. And passing NULL throws.
```
PHP Fatal error:  Uncaught TypeError: feof(): Argument #1 ($stream) must be of type resource, null given in /vendor/sabre/vobject/lib/Parser/MimeDir.php:283
Stack trace:
#0 /vendor/sabre/vobject/lib/Parser/MimeDir.php(283): feof(NULL)
#1 /vendor/sabre/vobject/lib/Parser/MimeDir.php(144): Sabre\VObject\Parser\MimeDir->readLine()
#2 /vendor/sabre/vobject/lib/Parser/MimeDir.php(90): Sabre\VObject\Parser\MimeDir->parseDocument()
#3 /vendor/sabre/vobject/lib/Reader.php(47): Sabre\VObject\Parser\MimeDir->parse(NULL, 3)
```

phil-davis:
@alecpl see PR #658 

Do you think that this is enough - throwing an exception to avoid the PHP fatal error?

I could have tried to detect this directly in `Reader::read()` - but it seemed to work by detecting the problem inside `parse` and throwing the exception there.

phil-davis:
Fixed by #658 
It will be in the next release "soon", and I will sort out back-porting the the various major release branches.

phil-davis:
Released in https://github.com/sabre-io/vobject/releases/tag/4.5.5

PR #658, opened 2024-05-27
throw ParseException when null input is provided

phil-davis:
Issue #657 

`parse` can be called with input `null`, but the caller should have previously provided an input stream that is already set up.

The change in this PR makes `parse` throw a `ParseException` if it is called the first time with `null`

This should help anyone calling this, or `Reader::read()` with no input. Instead of getting a PHP Fatal error, they will get an exception.

### DIFF
--- a/lib/Parser/MimeDir.php
+++ b/lib/Parser/MimeDir.php
@@ -79,6 +79,12 @@ class MimeDir extends Parser
             $this->setInput($input);
         }
 
+        if (!\is_resource($this->input)) {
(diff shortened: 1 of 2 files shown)


alecpl:
The `$input` parameter is marked as `string|resource|null`. So, this is a wrong solution.

alecpl:
I'm sorry, I was confused. The solution looks all right.

staabm (approved):

PR #664, opened 2024-08-01
Fix VTODO validation for PERCENT-COMPLETE

### LINKED ISSUE #663: VTODO PERCENT-COMPLETE property is not validated correctly
According to [RFC 5545](https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.8) the property to convey the percent completion of a to-do is called `PERCENT-COMPLETE`. 
However, in [`VTodo.php`](https://github.com/sabre-io/vobject/blob/1b388aab5cdbcb75a09be25f2e9119fc4463ae59/lib/Component/VTodo.php#L107) in the validation rules it checks for just `PERCENT`. This means that, for example, this VTODO does not trigger any errors when analyzed by the `validate` method, even though the `PERCENT-COMPLETE` property occurs two times which is invalid according to the standard.
```
BEGIN:VTODO
UID:8ed267e1-67c4-467d-8ae2-28e6ff03b033
DTSTAMP:20240729T133309Z
SUMMARY:TestTitle
COMPLETED;TZID=Europe/Berlin:20240729T133314
STATUS:COMPLETED
PERCENT-COMPLETE:70
PERCENT-COMPLETE:80
END:VTODO
```

Why is the library checking for `PERCENT`? Am I missing something?

jarikmarwede:
This PR should fix the linked issue I created about the `PERCENT-COMPLETE` property of the `VTODO` being misspelled as just `PERCENT`. I also added a test to verify that the validation of the property now works as expected. If you think the test is unnecessary, I can also just remove it again.

Fixes #663

### DIFF
--- a/lib/Component/VTodo.php
+++ b/lib/Component/VTodo.php
@@ -104,7 +104,7 @@ class VTodo extends VObject\Component
             'LAST-MODIFIED' => '?',
             'LOCATION' => '?',
             'ORGANIZER' => '?',
-            'PERCENT' => '?',
+            'PERCENT-COMPLETE' => '?',
             'PRIORITY' => '?',
             'RECURRENCE-ID' => '?',
             'SEQUENCE' => '?',
--- a/tests/VObject/Component/VTodoTest.php
+++ b/tests/VObject/Component/VTodoTest.php
@@ -168,4 +168,31 @@ HI;
             'DUE must occur after DTSTART',
         ], $messages);
     }
+
+    public function testValidateDuplicatePercentComplete(): void
+    {
+        $input = <<<HI
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:YoYo
+BEGIN:VTODO
+UID:8ed267e1-67c4-467d-8ae2-28e6ff03b033
+DTSTAMP:20240729T133309Z
+PERCENT-COMPLETE:70
+PERCENT-COMPLETE:80
+END:VTODO
+END:VCALENDAR
+HI;
+        $obj = Reader::read($input);
+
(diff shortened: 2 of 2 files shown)


phil-davis:
@staabm this PR changes "PERCENT" to "PERCENT-COMPLETE" (which is the correct name in the RFC spec)

If we merge this, then "PERCENT" will no longer be "officially" valid. But I suppose, if anybody has actually used it, "PERCENT" will just be an extra property that lives in some VTODO entries. It shouldn't break anything.

Do we want to also keep validating "PERCENT"?

phil-davis:
Also, maybe a separate issue to address, I don't see any code that checks that the value of PERCENT-COMPLETE is an integer from 1 to 100.

phil-davis (approved):
Code is fine, thanks. I would like to get the opinion of @staabm also about the old "PERCENT"

staabm:
I can't say much about it. this PR seems to align our impl with the RFC which feels like a good thing.

I don't know whether the previous impl is wrong though.

phil-davis:
> I don't know whether the previous impl is wrong though (or intentionally made like that for some unknown reason).

Actually, I don't think it really matters. For anyone currently using "PERCENT", it will still work. So they won't notice a difference.

For the clients that do the correct thing, and use "PERCENT-COMPLETE", they also will be fine, as long as they only specify it once, as per the spec.

staabm:
> Actually, I don't think it really matters. For anyone currently using "PERCENT", it will still work. So they won't notice a difference.

is this something we should test?

staabm:
thank you both!

PR #665, opened 2024-08-01
test: extra property is allowed in VTODO

phil-davis:
Follow-on from #664 

Tests that a now-unknown property called "PERCENT" is still accepted in a VTODO.


staabm (approved):
excellent!

PR #674, opened 2024-09-10
Allow unknown value data types for VALUE

heiglandreas:
Currently value-types that are not known will cause the parser to throw an InvalidDataException. In [RFC5545, Section 3.2.20](https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.20) is described though that the allowed values for the "VALUE" data-type can include x-names and iana-tokens. These can be any string that might - or might not - be understood by the parser. The description clearly states that "Applications MUST preserve the value data for x-name and iana-token value that they don't recognize without attempting to interpret or parse the value data"

This means that the content of the "VALUE" part should not be interpreted at all the moment the parser doesn't find a match in the corresponding mapping table.

But it should also not hard-fail when someone sets a VALUE that might not be understood. In such a case the default should be used instead.

### DIFF
--- a/lib/Document.php
+++ b/lib/Document.php
@@ -186,8 +186,8 @@ abstract class Document extends Component
         //  parameter need not be specified.  However, if the property's
         //  default value type is overridden by some other allowable value
         //  type, then this parameter MUST be specified.
-        if (isset($parameters['VALUE'])) {
-            $class = $this->getClassNameForPropertyValue($parameters['VALUE']);
+        if (!$valueType) {
+            $valueType = $parameters['VALUE'] ?? null;
         }
 
         if ($valueType) {
--- a/lib/Document.php
+++ b/lib/Document.php
@@ -180,23 +180,30 @@ abstract class Document extends Component
 
         $class = null;
 
+        // If a VALUE parameter is supplied, we have to use that
+        // According to https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.20
+        //  If the property's value is the default value type, then this
+        //  parameter need not be specified.  However, if the property's
+        //  default value type is overridden by some other allowable value
+        //  type, then this parameter MUST be specified.
+        if (isset($parameters['VALUE'])) {
+            $class = $this->getClassNameForPropertyValue($parameters['VALUE']);
+        }
+
         if ($valueType) {
             // The valueType argument comes first to figure out the correct
             // class.
             $class = $this->getClassNameForPropertyValue($valueType);
         }
 
+        // If the value parameter is not set or set to something we do not recognize
+        // we do not attempt to interpret or parse the datass value as specified in
+        // https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.20
+        // So when we so far did not get a class-name, we use the default for the property
         if (is_null($class)) {
-            // If a VALUE parameter is supplied, we should use that.
-            if (isset($parameters['VALUE'])) {
-                $class = $this->getClassNameForPropertyValue($parameters['VALUE']);
-                if (is_null($class)) {
-                    throw new InvalidDataException('Unsupported VALUE parameter for '.$name.' property. You supplied "'.$parameters['VALUE'].'"');
-                }
-            } else {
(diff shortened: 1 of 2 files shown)


lib/Document.php
```diff
@@ -180,23 +180,30 @@ public function createProperty(string $name, $value = null, ?array $parameters =
...
+        //  type, then this parameter MUST be specified.
+        if (isset($parameters['VALUE'])) {
+            $class = $this->getClassNameForPropertyValue($parameters['VALUE']);
+        }
+
         if ($valueType) {
             // The valueType argument comes first to figure out the correct
             // class.
             $class = $this->getClassNameForPropertyValue($valueType);
         }
```
phil-davis, line 197: If $valueType has been specified, and $parameters['VALUE'] has been set, then $class is calculated twice, the first result is immediately overwritten by the 2nd.

What am I not understanding?

phil-davis (commented):
Please add unit test cases to cover the changed/new behavior.
That will also make it easier to understand exactly what code-flow is needed for the various data combinations.

lib/Document.php
heiglandreas: No, that is correct. But now that you phrase it like that I might actually change it so that when `$valueType` evaluates to false and `$parameters['VALUE']` is set I will set it to `$parameters['VALUE']` instead. Then it would be truish in line 193 and the calculation only happens once.

Would that be a better approach?

phil-davis: > Would that be a better approach?

Yes, something like having an `else if` on the end of `if ($valueType) {`

phil-davis: The code you have done also works.

lib/Document.php
```diff
@@ -180,23 +180,30 @@ public function createProperty(string $name, $value = null, ?array $parameters =
...
+        // If the value parameter is not set or set to something we do not recognize
+        // we do not attempt to interpret or parse the datass value as specified in
+        // https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.20
+        // So when we so far did not get a class-name, we use the default for the property
         if (is_null($class)) {
-            // If a VALUE parameter is supplied, we should use that.
-            if (isset($parameters['VALUE'])) {
-                $class = $this->getClassNameForPropertyValue($parameters['VALUE']);
-                if (is_null($class)) {
-                    throw new InvalidDataException('Unsupported VALUE parameter for '.$name.' property. You supplied "'.$parameters['VALUE'].'"');
```
phil-davis, line 194: The code used to be able to throw this exception if `$parameters['VALUE']` had an invalid value (did not end up matching to a class).
Now, if `$parameters['VALUE']` is not valid, the code will end up doing:
`$class = $this->getClassNameForPropertyName($name);`

That is, it will just ignore the invalid value, and use what has been passed in `$name`.

phil-davis: I am surprised that no unit test has failed. I guess that there is not a unit test that covers the (now deleted) `InvalidDataException`

heiglandreas: Yes. As far as I see this part of [RFC5545, Section 3.2.20 (Value Data Type)](https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.20)

> Applications MUST preserve the value data for x-name and iana-token values that they don't recognize without attempting to interpret or parse the value data.

we need to allow to keep - but not interpret - VALUE information that we don't understand. That didn't work previously due to the Exception. That broke the whole process of interpretation. 

The Exception has been introduced with https://github.com/sabre-io/vobject/commit/c20acd7de21196aa829100aa3d4b44e213497c4f about 8 years ago to have "A better error message for unsupported VALUE parameters." but even then there should have been no Exception as unsupported VALUE parameters IMO do not pose an error condition. They should not be interpreted, yes (what they are now), but not break the process but also be preserved for other tools to interpret that might be able to understand it.

That is the reason why the current process then falls back to the default value when an "invalid" - let's rephrase that to "an uninterpretable" - value is encountered.

phil-davis (approved):

phil-davis:
Backport to 4.5 branch is PR #699

PR #691, opened 2024-10-01
fix: include RDATE in busy calculation

SebastianKrupinski:
fix: include RDATE in busy calculation

ICal spec allows for events with only an RDATE. At the moment if an event only includes an RDATE and no RRULE it is skipped due to the current code path only checking for the existence of a RRULE.

RFC Reference https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.5.2

### DIFF
--- a/lib/FreeBusyGenerator.php
+++ b/lib/FreeBusyGenerator.php
@@ -359,7 +359,7 @@ class FreeBusyGenerator
 
                         $times = [];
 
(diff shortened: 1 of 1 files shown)


phil-davis:
@SebastianKrupinski can you add some unit test case(s) that demonstrate that RDATE specifications will now be correctly processed.

PR #692, opened 2024-10-11
Throw InvalidDataException when RRule is invalid

mathroc:


phil-davis:
Released in https://github.com/sabre-io/vobject/releases/tag/4.5.6

PR #694, opened 2024-10-14
chore: add PHP 8.4 to CI of master

phil-davis:
similar CI changes to what has been done in other sabre repos.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,12 +12,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.4', '8.0', '8.1', '8.2', '8.3']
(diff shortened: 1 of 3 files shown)


phil-davis:
php-cs-fixer passes in PHP 8.4 with the added settings for the `nullable_type_declaration` rule. So all of those PHP 8.4 deprecation warnings have been fixed.

.github/workflows/ci.yml
```diff
@@ -12,12 +12,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.4', '8.0', '8.1', '8.2', '8.3']
```
phil-davis, line 15: Note: 7.4 is already in the `include` section, so does not need to be duplicated here.

PR #707, opened 2024-12-05
add phpdoc return types for arrayaccess methods (fixes #563)

### LINKED ISSUE #563: Deprecation warnings on PHP 8.1
Hello. I'm using PHP 8.1.1 and lib version 4.4.1

I see the following deprecation warnings
```
1x: Method "ArrayAccess::offsetSet()" might add "void" as a native return type declaration in the future. Do the same in implementation "Sabre\VObject\Node" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "ArrayAccess::offsetUnset()" might add "void" as a native return type declaration in the future. Do the same in implementation "Sabre\VObject\Node" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "ArrayAccess::offsetSet()" might add "void" as a native return type declaration in the future. Do the same in implementation "Sabre\VObject\Property" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "ArrayAccess::offsetUnset()" might add "void" as a native return type declaration in the future. Do the same in implementation "Sabre\VObject\Property" now to avoid errors or add an explicit @return annotation to suppress this message.

  1x: Method "ArrayAccess::offsetSet()" might add "void" as a native return type declaration in the future. Do the same in implementation "Sabre\VObject\Property\ICalendar\DateTime" now to avoid errors or add an explicit @return annotation to suppress this message.
```

smhg:
See #563.

This adds only phpdoc return types as real return types would not work for the earlier php versions this library supports.

### DIFF
--- a/lib/ElementList.php
+++ b/lib/ElementList.php
@@ -24,6 +24,7 @@ class ElementList extends ArrayIterator
      *
      * @param int   $offset
      * @param mixed $value
+     * @return void
      */
     #[\ReturnTypeWillChange]
     public function offsetSet($offset, $value)
@@ -37,6 +38,7 @@ class ElementList extends ArrayIterator
      * This method just forwards the request to the inner iterator
      *
      * @param int $offset
+     * @return void
      */
     #[\ReturnTypeWillChange]
     public function offsetUnset($offset)
--- a/lib/ElementList.php
+++ b/lib/ElementList.php
@@ -24,6 +24,7 @@ class ElementList extends ArrayIterator
      *
      * @param int   $offset
      * @param mixed $value
+     *
      * @return void
      */
     #[\ReturnTypeWillChange]
@@ -38,6 +39,7 @@ class ElementList extends ArrayIterator
      * This method just forwards the request to the inner iterator
      *
      * @param int $offset
+     *
      * @return void
      */
     #[\ReturnTypeWillChange]
--- a/lib/Node.php
+++ b/lib/Node.php
@@ -212,6 +212,7 @@ abstract class Node implements \IteratorAggregate, \ArrayAccess, \Countable, \Js
      *
      * @param int   $offset
      * @param mixed $value
+     *
      * @return void
      */
     #[\ReturnTypeWillChange]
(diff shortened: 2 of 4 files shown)


phil-davis:
@smhg php-cs-fixer wants an extra comment line between `@param` and `@return` PHPdoc lines.
Can you adjust that please?

And this will not change run-time behavior. I suppose that the messages reported in issue #563 are being generated by the analysis tools, and not run-time in production?

smhg:
@phil-davis I've added the extra lines to the phpdoc blocks. I'm sorry, trusted CI would run php-cs-fixer as it requires an older PHP env.

The warnings in #563 are shown by PHPUnit only (v9 in my case). I tried to make the minimal required change to close the issue. I was unsure adding a real return value would not add any complexities on older supported PHP versions for instance. I reasoned it should only be a patch release.

phil-davis (approved):

phil-davis:
I will have a look at what else is waiting for 4.5 branch and hopefully make the next patch release later today.

smhg:
Thanks @phil-davis!

Now I also remember I tried to bump phpstan's level to detect these deprecation warnings, but I realized that is probably what's going on in the main branch? That's not worth the effort in v4 I assumed?

phil-davis:
> Now I also remember I tried to bump phpstan's level to detect these deprecation warnings, but I realized that is probably what's going on in the main branch? That's not worth the effort in v4 I assumed?

Yes, I was making progress on increasing the phpstan level in the master branch. I did it easily in the smaller sabre-io repos, but it is a lot more effort here in vobject, so I lost enthusiasm! But some day I would like that to get finished, and release it as a new major version (because it is likely that explicitly declaring parameter and return types will result in some technical backward-compatibility breaks)

So contributions for phpstan enhancements are welcome and go to the master branch.

smhg:
FWIW I now noticed a package called `symfony/phpunit-bridge` detects and adds these deprecation notices in (my) projects. I think this merge doesn't harm anyone, but it also doesn't require a release I guess.

I'm sorry about the distraction. I'll add a message how to disable these in the original issue.

PR #715, opened 2025-06-02
Handle different sequences in base event and exception event

gharlan:
The sequence in exception events can be different to the sequence of the base event. But the ITip broker always replies with the base sequence number, even for replies to exception events.

The pull requests shows the bug.

From iCalender RFC:
> Recurrence instances of a recurring component MAY have different sequence numbers.

https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.7.4

Google Calendar uses different sequence numbers for exception events.  
Example: Base event has sequence 2. When creating a new exception event, it gets the sequence number 3, but the base event keeps number 2.
Now when responding to the exception event via VObject iTip broker, the response contains the wrong sequence number 2, so the response is ignored by Google calender.


gharlan:
Added a fix now.

Issue #700, opened 2024-10-29
RRULE with RDATE / EventIterator.php


munderl:
I think there is a problem with the EventIterator.php. In rfc5545 a "RDATE" can occur inside a "RRULE". In such a case, the recurrence instances are defined by the union of occurrences defined by both the "RDATE" and "RRULE". In the current implementation, the "RRULE" is ignored completely when an additional "RDATE" is included in the "VEVENT". Therefore the EventIterator returns only the "RDATE" events and the original "RRULE" is lost.

munderl:
Seems as somebody in 2016 already mentioned this: [#347](https://github.com/sabre-io/vobject/issues/347). I encountered such a vevent with RRULE and RDATE several time now, so it seems to be used. Might be created by either thunderbird or DAVx^5.

Issue #347, opened 2016-08-25
Composing multiple RDATE and RRULE properties


brianmuse:
I could be mistaken, but I believe that RFC5545 specifies that multiple `RDATE` properties may exist along with `RRULE`.

For example, this would be valid and would mean it recurs yearly and on those `RDATE` values:

```
RDATE:20160607T000000Z
RDATE:20160707T000000Z
RRULE:FREQ=YEARLY
```

Also, though it has `SHOULD NOT` guidance, you may also encounter multiple `RRULE` properties:

```
RDATE:20160607T000000Z
RDATE:20160707T000000Z
RRULE:FREQ=YEARLY
RRULE:FREQ=MONTHLY
```

It appears that both cannot be considered at the same time and that multiple are not supported:
https://github.com/fruux/sabre-vobject/blob/master/lib/Recur/EventIterator.php#L181-L190

```PHP
        if (isset($this->masterEvent->RDATE)) {
            $this->recurIterator = new RDateIterator(
                $this->masterEvent->RDATE->getParts(),
                $this->startDate
            );
        } elseif (isset($this->masterEvent->RRULE)) {
            $this->recurIterator = new RRuleIterator(
                $this->masterEvent->RRULE->getParts(),
                $this->startDate
            );
        }
```

We have implemented this behavior before on another project and our technique (even supporting `EXRULE`) was to create an iterator for each `RRULE`, `EXRULE` and an iterator for `RDATES` and `EXDATES` and then compose result from iterating over all four of those one element at a time to get the "next" result (using logic to determine which event was actually next while handling any exceptions from `EXRULE` or `EXDATE`).

We used generators with PHP5.6, however iterators may be just as good in this case.

evert:
Hi Brian,

I agree that this is definitely a real iCalendar feature we don't support. Most of the focus has been on trying to get support for the most common cases, and I haven't really need mixed RDATE/RRULE yet in the wild. Have you?

brianmuse:
Thanks for the prompt reply! I can appreciate the focus on the most impactful areas.

We primarily work with the Google Calendar API where we have encountered this situation, though rare, for a few of our customers. I'm not 100% sure if they're able to someone create that state through and of the Google Calendar GUIs or whether its through another integration with Google Calendar from another schedule service or because of direct use of the Google Calendar API (which does allow this). 

It's been imperative for our use case to not miss anything during instance expansion, so despite it being a bit of an edge it's still something that we have to handle. 

[...]

PR #716, opened 2025-06-20
Support concurrency of multiple RDATE together with RRULE directives.

rotdrop:
This PR address #347 and its duplicate #700. The commit also adds a test. The idea is to maintain an array of iterators, compare their current values and pick the smallest one in next(). Then all iterators matching the timestamp value of the chosen DateTime are advanced, which the others which may be invalid or have their individual current values still ahead of time are kept at their position.

I have added a test (with one RRULE and RDATES which are placed "out of order").

### DIFF
--- a/lib/Recur/EventIterator.php
+++ b/lib/Recur/EventIterator.php
@@ -163,18 +163,29 @@ class EventIterator implements \Iterator
             $this->eventDuration = 0;
         }
 
+        $this->recurIterators = [];
+        $isRecurring = false;
+        if (isset($this->masterEvent->RRULE)) {
+            foreach ($this->masterEvent->RRULE as $rRule) {
+                $this->recurIterators[] = new RRuleIterator(
+                    $this->masterEvent->RRULE->getParts(),
+                    $this->startDate
+                );
+            }
+            $isRecurring = true;
+        }
         if (isset($this->masterEvent->RDATE)) {
-            $this->recurIterator = new RDateIterator(
-                $this->masterEvent->RDATE->getParts(),
-                $this->startDate
-            );
-        } elseif (isset($this->masterEvent->RRULE)) {
-            $this->recurIterator = new RRuleIterator(
-                $this->masterEvent->RRULE->getParts(),
-                $this->startDate
-            );
-        } else {
-            $this->recurIterator = new RRuleIterator(
+            foreach ($this->masterEvent->RDATE as $rDate) {
+                $this->recurIterators[] = new RDateIterator(
+                    $rDate->getParts(),
+                    $this->startDate,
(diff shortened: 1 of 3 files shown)


rotdrop:
Actually, the changes to the `RDateIterator.php` would not be needed; I started with the PHP AppendIterator, but as this was not enough: duplicated event dates are handled by the EventIterator anyway, which now behaves like an "OrderedUniqueAppendIterator".

rotdrop:
> Actually, the changes to the `RDateIterator.php` would not be needed; I started with the PHP AppendIterator, but as this was not enough: duplicated event dates are handled by the EventIterator anyway, which now behaves like an "OrderedUniqueAppendIterator".

Nope, it ain't that easy. However, this is also due to the current code base. If the current event DTSTART refers to a date after the dates defined by an RDATE rule, then the ordering of the values returned by the *current* `EventIterator` is garbled: it will always return the DTSTART value of the master event as first item *which is wrong*. RDATE exception may very well refer to events which are in the past w.r.t. DTSTART of the master event.

This also means that further changes would be needed. The RDate iterator has to take care of this.

lib/Recur/EventIterator.php
```diff
@@ -163,18 +163,26 @@ public function __construct($input, ?string $uid = null, ?\DateTimeZone $timeZon
...
-                $this->startDate
-            );
-        } else {
-            $this->recurIterator = new RRuleIterator(
+            foreach ($this->masterEvent->RDATE as $rDate) {
+                $this->recurIterators[] = new RDateIterator(
+                    $rDate->getParts(),
+                    $this->startDate,
+                );
+            }
```
SebastianKrupinski: Instead of creating an array of iterators would it be easier to combine all the rdate instances in to a single iterator?

```
$rdateValues = [];
foreach ($this->masterEvent->RDATE as $rdate) {
    $rdateValues = array_merge($rdateValues, $rdate->getParts());
}
$this->recurIterator = new RDateIterator(
    $this->masterEvent->RDATE->getParts(),
    $rdateValues,
    $this->startDate
);
```

This would reduce the complexity of iterating through RRULE and RDATE significantly, you would only need to rewind, forward, and check 2 iterators, instead of looping through an array of iterators.

rotdrop: Did not look into this for a long time now. I am no longer using upstream Sabre, but rather my own forks with my bug fixes, due to lack of responsiveness of the maintainers of Sabre. No flame intended, but we all have only a limited amount of time available, this applies as well to the maintainers of a software package as to potential contributors (which might be at the same time over-loaded maintainers of other software packages). I'll post a more useful response as soon as I find the time to look into your suggestions.

Issue #717, opened 2025-06-24
Get a property by multiple types


JimKnoxx:
I noticed that the VCard class only has a getByType method.
If the VCard contains multiple types for a property there is currently no easy way to select this exact property. 
If we have a VCARD like this:
~~~
BEGIN:VCARD
VERSION:3.0
TEL;TYPE=HOME,CELL:112233445566
TEL;TYPE=WORK,CELL:665544332211
TEL;TYPE=WORK:7778889994455
TEL;TYPE=EXTERNAL:555555555
END:VCARD
~~~

Selecting WORK will always return "665544332211".

PR #718, opened 2025-06-24
feat(VCard): add getByTypes method (#717)

JimKnoxx:
In this PR I added a getByTypes method that gets an string array to select an exact property of a VCard.

It checks case insensitive and need to exactly match the types of the wanted property.

### DIFF
--- a/lib/Component/VCard.php
+++ b/lib/Component/VCard.php
@@ -410,6 +410,34 @@ class VCard extends VObject\Document
     }
 
     /**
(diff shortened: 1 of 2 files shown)


phil-davis:
@staabm what do you think about add this method?

staabm (commented):
I don't see a reason to not include it in case you think its useful

phil-davis (approved):

Issue #722, opened 2025-08-28
Broker creates EXDATE property with floating timezone


KristofferFM:
If an organizer removes an attendee from a single instance of an recurrent event, and that event did not have an EXDATE property beforehand, Broker will add an EXDATE property with no timezone (i.e. floating timezone).
I believe this is wrong and should be fixed.

PR #723, opened 2025-08-28
Broker shouldn't add EXDATE property with floating timezone

KristofferFM:
Fixes issue #722


phil-davis (approved):
LGTM, thanks for the fix and tests.
I rebased so that the other code-formatting changes would go away - I merged the same format changes in another PR today.

PR #756, opened 2026-04-13
fix: Update deprecated timezone names

ralflang:
Supersedes #755 which was based against the current maintenance branch and thus had a huge, misleading diff.

fix(timezonedata): Replace deprecated IANA timezone names

Systems without the tzdata-legacy package, i.e. (Ubuntu 24.04+) reject deprecated names.

Examples: Europe/Kiev, Asia/Calcutta, America/Godthab
This causes FindFromTimezoneMap::find() to throw.
Update all 13 occurrences across windowszones, lotuszones and exchangezones maps to their canonical replacements and wrap DateTimeZone instantiation in try/catch for resilience against future renames.

Comes with unit test


lib/timezonedata/windowszones.php
```diff
@@ -82,10 +82,10 @@
     'Morocco Standard Time' => 'Africa/Casablanca',
     'Mountain Standard Time' => 'America/Denver',
     'Mountain Standard Time (Mexico)' => 'America/Chihuahua',
-    'Myanmar Standard Time' => 'Asia/Rangoon',
+    'Myanmar Standard Time' => 'Asia/Yangon',
     'N. Central Asia Standard Time' => 'Asia/Novosibirsk',
     'Namibia Standard Time' => 'Africa/Windhoek',
-    'Nepal Standard Time' => 'Asia/Katmandu',
+    'Nepal Standard Time' => 'Asia/Kathmandu',
```
phil-davis, line 88: Yes! This has been annoying having Kathmandu spelled wrong.

phil-davis (approved):
This LGTM.
@staabm what do you think?
Can you see any issue about the old names etc?

lib/TimezoneGuesser/FindFromTimezoneMap.php
```diff
@@ -22,7 +22,11 @@ public function find(string $tzid, ?bool $failIfUncertain = false): ?\DateTimeZo
     {
         // Next, we check if the tzid is somewhere in our tzid map.
         if ($this->hasTzInMap($tzid)) {
-            return new \DateTimeZone($this->getTzFromMap($tzid));
```
staabm, line 25: before this PR the observed behaviour of `find` was that it would throw (even if it was not a documented behaviour).

I find the new behaviour more useful, but I think we need to at least document this "breaking change" - we can use a `@throws void` phpdoc (for static analysis support), but can also document it in the changelog

phil-davis: done, and will mention this in the changelog/release notes. This can be part of a minor version bump.

staabm (approved):

Issue #411, opened 2018-04-25
FreeBusy Period serialization to JSON removes UTC indicator (Z)


renaudboyer:
When serializing FreeBusy Period objects to JSON, UTC Z is stripped from date. And Json dates are in local time instead of UTC.

Here's the iCal (headers have been removed)

```
BEGIN:VCALENDAR
BEGIN:VFREEBUSY
DTSTART:20180302T000000Z
DTEND:20180307T000000Z
DTSTAMP:20180418T074609Z
FREEBUSY:20180302T100000Z/20180302T110000Z
END:VFREEBUSY
END:VCALENDAR
```

is serialized to

```
[
  "vcalendar",
  [
    [
      "vfreebusy",
      [
        [
          "dtstart", {},
          "date-time", "2018-03-02T00:00:00Z"
        ],
        [
          "dtend", {},
          "date-time", "2018-03-07T00:00:00Z"
        ],
        [
          "dtstamp", {},
          "date-time", "2018-04-18T07:57:26Z"
        ],
        [
          "freebusy", {},
          "period",
          [
            "2018-03-02T10:00:00",
            "2018-03-02T11:00:00"
          ]
        ]
      ],
      []
    ]
  ]
]
```

DominikTo:
Is this the same object? In your iCalendar document `DTSTAMP` is 20180418T0**74609**Z. In the JSON document it's 2018-04-18T0**7:57:26**Z.

I only had a very quick look. If I understand [L290 here](https://github.com/sabre-io/vobject/blob/4.1.6/lib/Property/ICalendar/DateTime.php#L290) correctly, we are adding the Z in [L296](https://github.com/sabre-io/vobject/blob/4.1.6/lib/Property/ICalendar/DateTime.php#L296) if all of the following are true:

* `isFloating()` is `false`
* `$isUtc` is `true` (it's true if timezone is UTC, GMT or Z)

But on first glance I am also confused, why the Free-Busy Info has no TZ info.
Maybe @evert has an idea?

evert:
I'm not sure why it's happening, but I _can_ reproduce it

bastien-roucaries:
Hi added pull request

PR #765, opened 2026-05-31
feat: preserve UTC 'Z' indicator in Period JSON serialization

bastien-roucaries:
When serializing PERIOD values to JSON (for jCal format), preserve the 'Z' timezone indicator for UTC datetimes. This ensures FREEBUSY periods maintain explicit UTC designation, avoiding timezone ambiguity.

Without this change:
  FREEBUSY:20120226T230000Z/20120226T230000Z
  => serializes to: ["2012-02-26T23:00:00", "2012-02-26T23:00:00"]

With this change:
  FREEBUSY:20120226T230000Z/20120226T230000Z
  => serializes to: ["2012-02-26T23:00:00Z", "2012-02-26T23:00:00Z"]

This behavior is consistent with RFC 5545 which requires UTC times to be designated with the 'Z' suffix.

Related: https://github.com/sabre-io/vobject/issues/411

### DIFF
--- a/lib/Property/ICalendar/Period.php
+++ b/lib/Property/ICalendar/Period.php
@@ -70,6 +70,11 @@ class Period extends Property
         parent::setJsonValue($value);
     }
 
+    public function appendUtc(string $strDate)
+    {
+        return !str_ends_with($strDate, 'Z') ? '' : 'Z';
+    }
+
     /**
      * Returns the value, in the format it should be encoded for json.
      *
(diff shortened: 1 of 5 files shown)


lib/Property/ICalendar/Period.php
```diff
@@ -87,17 +87,21 @@ public function getJsonValue(): array
...
 
+            // Check if the datetime is in UTC timezone
+            $isUtc = in_array($start->getTimezone()->getName(), ['UTC', 'GMT', 'Z', '+00:00']);
+            $zSuffix = $isUtc ? 'Z' : '';
+
             // This is a duration value.
             if ('P' === $end[0]) {
                 $return[] = [
-                    $start->format('Y-m-d\\TH:i:s'),
+                    $start->format('Y-m-d\\TH:i:s').$zSuffix,
```
phil-davis: In this case the "date"-time string is a duration, so a timezone indication is not relevant.
Something like "1970-01-02:14:30:00" just indicates a duration of 1 day, 14 hours, 30 miutes and 0 seconds.
Or am I not understanding?

phil-davis (commented):
Please add/edit unit test cases do demonstrate what is expected, and that it works.
Thanks

lib/Property/ICalendar/Period.php
bastien-roucaries: @phil-davis for end, not for start that should use timezone

phil-davis:
@staabm @Chris53897 @bastien-roucaries I adjusted the code to append the "Z" when the input ends with "Z" (rather than just contains "Z"). I think that the only valid place that "Z" can be is at the end of a date-time string.

I added some test cases to DateTime `testValidate` to check that things like `20260627T100000Z+0200` are already being seen as invalid.

Please review and give your opinion.

bastien-roucaries:
@phil-davis I think iut is safe

phil-davis:
Rebased and squashed. The previous commit history was mostly "rubbish" tinkering getting code-style etc passing.

PR #767, opened 2026-06-11
Drop PHP 7

phil-davis:
- drop PHP 7.4 8.0 and 8.1 support
- apply cs-fixer and rector automated changes
- make changes to test code and real code to get phpstan to pass

### DIFF
--- a/lib/Document.php
+++ b/lib/Document.php
@@ -118,10 +118,10 @@ abstract class Document extends Component
     public function create(string $name)
     {
         if (isset(static::$componentMap[strtoupper($name)])) {
-            return call_user_func_array([$this, 'createComponent'], func_get_args());
+            return call_user_func_array($this->createComponent(...), func_get_args());
         }
 
-        return call_user_func_array([$this, 'createProperty'], func_get_args());
+        return call_user_func_array($this->createProperty(...), func_get_args());
     }
 
     /**
--- a/lib/Document.php
+++ b/lib/Document.php
@@ -118,10 +118,10 @@ abstract class Document extends Component
     public function create(string $name)
     {
         if (isset(static::$componentMap[strtoupper($name)])) {
-            return call_user_func_array($this->createComponent(...), func_get_args());
+            return $this->createComponent(...func_get_args());
         }
 
-        return call_user_func_array($this->createProperty(...), func_get_args());
+        return $this->createProperty(...func_get_args());
     }
 
     /**
--- a/lib/PHPUnitAssertions.php
+++ b/lib/PHPUnitAssertions.php
@@ -55,13 +55,13 @@ trait PHPUnitAssertions
         $actual = $getObj($actual)->serialize();
 
         // Finding wildcards in expected.
-        preg_match_all('|^([A-Z]+):\\*\\*ANY\\*\\*\r$|m', $expected, $matches, PREG_SET_ORDER);
+        preg_match_all('|^([A-Z]+):\\*\\*ANY\\*\\*\r$|m', (string) $expected, $matches, PREG_SET_ORDER);
 
         foreach ($matches as $match) {
             $actual = preg_replace(
                 '|^'.preg_quote($match[1], '|').':(.*)\r$|m',
                 $match[1].':**ANY**'."\r",
-                $actual
+                (string) $actual
             );
         }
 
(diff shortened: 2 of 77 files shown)


phil-davis:
```
fopen(foo.json): Failed to open stream: No such file or directory
```

I have no idea how the tests found `foo.json` previous to the changes in this PR. It doesn't seem to exist as a fixture anywhere in this repo. The tests are broken now, but they passed in master that has PHPunit9. This has PHPunit11. No idea if that is relevant. I get the same stuff failing locally on my laptop.

phil-davis:
`git log --all --full-history -- "*foo*"`

Finds nothing. So there never has been any file in this repo with "foo" in its name. And I don't see any code that will create `foo.json` or `foo.ics`. So how did this ever work?

Edit add:

The test cases actually expected the Cli to exit with status "2". So they did expect that a problem would be reported - the input file does not exist. Due to changes in PHPunit11, the test code was now exposed to the PHP warning from fopen. So I needed to handle that in the test code, and make a small addition to the real code to throw an InvalidDataException when `fopen` returns false. That is then handled causing `exit 2`, the previous behavior.

lib/Document.php
```diff
@@ -118,10 +118,10 @@ public function getDocumentType(): int
     public function create(string $name)
     {
         if (isset(static::$componentMap[strtoupper($name)])) {
-            return call_user_func_array([$this, 'createComponent'], func_get_args());
+            return call_user_func_array($this->createComponent(...), func_get_args());
```
staabm: can this be simplified?

```suggestion
            return$this->createComponent(func_get_args());
```

same for the line below

tests/VObject/CliTest.php
```diff
@@ -154,7 +154,7 @@ public function testConvertJson(): void
     public function testConvertJCardPretty(): void
     {
         if (version_compare(PHP_VERSION, '5.4.0') < 0) {
-            $this->markTestSkipped('This test required PHP 5.4.0');
+            $this::markTestSkipped('This test required PHP 5.4.0');
```
staabm: should be `self::` instead of `$this::`

phil-davis: Fixed in commit 99d877e13485f8a5d6c12b821d7904856a191899

lib/Document.php
phil-davis: Simplified in https://github.com/sabre-io/vobject/pull/767/changes/2d35ee74eca8d236d4c175fdade91cc73b527b24

marclaporte:
Will this lead to a new major release? (does dropping old PHP versions trigger SemVer?)

phil-davis:
> Will this lead to a new major release? (does dropping old PHP versions trigger SemVer?)

Normally I would just make a minor release for dropping PHP versions. `composer` looks after knowing which version suits the PHP version(s) supported by a project.
But there are also a lot of changes/enhancements in `master` that have parameter and return type definitions, and other PHPdoc stuff. So I think that consumers might find it a bit of effort to implement, particularly if they are currently calling methods with wrong parameter types that have been coerced by PHP, but now the types will be strictly enforced.

So I will make this a major version. That gives consumers the option to specifically choose to update.

lib/PHPUnitAssertions.php
```diff
@@ -51,23 +51,26 @@ public function assertVObjectEqualsVObject($expected, $actual, string $message =
...
-        $actual = $getObj($actual)->serialize();
+        /**
+         * @var string $expectedSerialized
+         */
+        $expectedSerialized = $getObj($expected)->serialize();
+        $actualSerialized = $getObj($actual)->serialize();
 
         // Finding wildcards in expected.
-        preg_match_all('|^([A-Z]+):\\*\\*ANY\\*\\*\r$|m', $expected, $matches, PREG_SET_ORDER);
+        preg_match_all('|^([A-Z]+):\\*\\*ANY\\*\\*\r$|m', $expectedSerialized, $matches, PREG_SET_ORDER);
```
staabm: This is missing handling if a possible error return

lib/Property.php
```diff
@@ -118,12 +99,12 @@ public function setValue($value): void
      * multi-value object, some decision will be made first on how to represent
      * it as a string.
      *
-     * To get the correct multi-value version, use getParts.
+     * To get the corgetValuerect multi-value version, use getParts.
```
staabm: Typo?

staabm (approved):

lib/Property.php
phil-davis: Well spotted. Random paste of the function name below! Fixed.

lib/PHPUnitAssertions.php
phil-davis: Added a check of the return result of `preg_match_all`

PR #770, opened 2026-06-15
chore: prepare major release 5.0.0

phil-davis:
5.0.0 (2026-06-15)
------------------
* #767 Drop PHP 7.4 8.0 8.1 and support PHP 8.2 and up (@phil-davis)
* #769 fix: minor code fixes (@phil-davis)
* #586 PHP types (@tcitworld @phil-davis)

This major release v5 declares function parameter and return types across the
codebase. Users will need to be careful to call functions/methods with the
correct types.

Support for PHP below 8.2 has been dropped.
This release supports PHP 8.2 and up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,19 @@
 ChangeLog
 =========
 
+5.0.0 (2026-07-07)
+------------------
+* #767 Drop PHP 7.4 8.0 8.1 and support PHP 8.2 and up (@phil-davis)
+* #769 fix: minor code fixes (@phil-davis)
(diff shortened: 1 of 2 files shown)


phil-davis:
Notes:
We haven't increased the `phpstan` level here. There could be lots of changes to do, such as #498 (that has been sitting there for years). Or adding array shape PHPdoc throughout, and so on. But that all requires quite a bit of effort.

We have applied `rector` to the code, and having all the parameter and return types is "a good thing".

I am proposing to make this major version 5, so that everyone can decide when they want to upgrade.

PR #771, opened 2026-06-20
chore: refactor to declare more types

phil-davis:
Trying some changes to type declarations now that PHP 8.2 is the minimum.

I thought that it might be easy to remove lots of `ReturnTypeWillChange`.
But it turns out to not always be simple. See a few comments below.

### DIFF
--- a/lib/Node.php
+++ b/lib/Node.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Sabre\VObject;
 
 use Sabre\Xml;
@@ -62,17 +64,12 @@ abstract class Node implements \IteratorAggregate, \ArrayAccess, \Countable, \Js
     /**
      * This method returns an array, with the representation as it should be
      * encoded in JSON. This is used to create jCard or jCal documents.
-     *
-     * @return array|string
      */
-    #[\ReturnTypeWillChange]
-    abstract public function jsonSerialize();
+    abstract public function jsonSerialize(): array|string|null;
 
     /**
      * This method serializes the data into XML. This is used to create xCard or
      * xCal documents.
-     *
-     * @param Xml\Writer $writer XML writer
      */
     abstract public function xmlSerialize(Xml\Writer $writer): void;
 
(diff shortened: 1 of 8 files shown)


lib/Node.php
```diff
@@ -159,11 +154,8 @@ public function count(): int
      * Checks if an item exists through ArrayAccess.
      *
      * This method just forwards the request to the inner iterator
-     *
-     * @param int $offset
      */
-    #[\ReturnTypeWillChange]
-    public function offsetExists($offset): bool
+    public function offsetExists(mixed $offset): bool
```
phil-davis, line 158: These don't look so good. The PHPdoc used to say that `$offset` is an `int`
But the parent ArrayAccess::offsetExists(mixed $offset): bool` allows anything as parameter 1.
So we can't write `public function offsetExists(int $offset): bool`
So we seem to lose something here, the PHPdoc type hint goes, but we can't preserve it in the actual parameter type declaration.

tests/VObject/ParameterTest.php
```diff
@@ -30,17 +32,17 @@ public function testModify(): void
         $cal = new Component\VCalendar();
 
         $param = new Parameter($cal, 'name', null);
-        $param->addValue(1);
+        $param->addValue('1');
```
phil-davis, line 35: IMO `addValue` was supposed to always expect to be passed a string.
In this PR I enforced that. But this test was passing `int` literals to `addValue`

CarlSchwan (approved):

lib/Parameter.php
```diff
@@ -88,10 +88,8 @@ public static function guessParameterNameByValue(string $value): string
      * Updates the current value.
      *
      * This may be either a single, or multiple strings in an array.
```
CarlSchwan, line 90:
```suggestion
     * This may be either a single, or multiple strings in an array.
     * @param string|list<string> $value
```

I would keep the type just make it more correct

lib/Parameter.php
```diff
@@ -32,19 +34,15 @@ class Parameter extends Node implements \Stringable
...
-     *
-     * @var string|array|null
      */
-    protected $value;
+    protected array|string|null $value = null;
 
     /**
      * Sets up the object.
      *
      * It's recommended to use the create:: factory method instead.
```
CarlSchwan, line 43:
```suggestion
     * It's recommended to use the create:: factory method instead.
     * 
     * @param string|list<string>|null $value
```

lib/Parameter.php
```diff
@@ -32,19 +34,15 @@ class Parameter extends Node implements \Stringable
 
     /**
      * Parameter value.
```
CarlSchwan, line 36:
```suggestion
     * Parameter value.
     * 
     * @var string|list<string>|null
```

PR #784, opened 2026-08-17
fix(iTip): Prevent single-occurrence reply to be a `significantChange` for everyone

DerDreschner:
Under some circumstances (e.g., replying to a single event in a recurring series with a lot of participants), all participants get a new `REQUEST` message delivered. It's pretty confusing for them, as this shouldn't happen. This is being fixed with this PR.

### DIFF
--- a/lib/ITip/Broker.php
+++ b/lib/ITip/Broker.php
@@ -491,6 +491,8 @@ class Broker
             }
         }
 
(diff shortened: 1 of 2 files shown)


DerDreschner:
CC @SebastianKrupinski @provokateurin

PR #789, opened 2026-09-09
Do not escape commas in Uri values

KristofferFM:
Uri values, as found in PHOTO and GEO properties, should not have comma escaped. An example of a valid PHOTO property can be seen here: https://datatracker.ietf.org/doc/html/rfc6350#section-6.2.4.

As a result of this change, I had to modify an existing testcase that uses a Uri value with geo: scheme.


KristofferFM:
I will look into the failing checks asap

PR #790, opened 2026-09-14
fix: handle UTC offset timezones

SebastianKrupinski:
### Summary
- added logic to handle UTC offset timezones properly "DTSTART;TZID="UTC-04:00":20260728T111500"


pi0n00r:
Thanks for moving this into the shared timezone parser. This covers the original `UTC-04:00` case (and the compact `UTC-0400` form) at the correct layer, so it is preferable to retaining a Calendar Dashboard-only correction.

The current CI failures appear limited to the two new PHPUnit data providers: PHPUnit 11 requires them to be static, and Rector expects `#[DataProvider(...)]` attributes rather than `@dataProvider` annotations. Once that is green, we can validate the dependency through Nextcloud and retire the application-local patch when the fixed vobject release is consumed.

pi0n00r:
Confirmed: the corrected head is green across PHP 8.2 through 8.5, including Rector and the fixed-offset regression. Thanks for turning this around so quickly.

CarlSchwan (approved):

